PR #19, opened 2022-02-06
Correcting a small typo in the badges section

BrightDV:
Hello,

Just a small pr to correct a typo in the badges section, which I found while translating the strings. 
Also, in a few days, I will open another pr to add a French translation of the app.

Thanks for this great launcher!


MM2-0:
Thank you!
Let me know if you need help with the translation, I think the strings aren't very well documented.

BrightDV:
>Let me know if you need help with the translation, I think the strings aren't very well documented.

For the moment, it's alright! But I found some duplicates in the strings, in particular the name of companies, e.g., Google, Nextcloud…
I will open a pr to remove them if necessary.
Also, is there a way in the app to select a language in particular, or is it using the system's default?

MM2-0:
It's using the system's default language

PR #24, opened 2022-02-08
French support for the app

BrightDV:
Hello,

Finally, this is the French translation of the app.
If I find later some spelling mistakes or improvements, I will open another PR!

Also, setting up a [Weblate](https://weblate.org/) or something like that would make translation of the app easier.


MM2-0:
Tanks a lot!

> Also, setting up a [Weblate](https://weblate.org/) or something like that would make translation of the app easier.
I know, I have already set up a Weblate instance but I haven't imported the project yet because I'd like to clean up some things and add some documentation before I do.

MM2-0:
@BrightDV 
I have setup a  Weblate instance at https://i18n.mm20.de.

I have never worked with Weblate before so I would appreciate if you could have a look (there are also 7 new strings that need translation).

BrightDV:
> I have never worked with Weblate before so I would appreciate if you could have a look (there are also 7 new strings that need translation).

I checked the instance, and it looks good to me!
I will open a PR to add Weblate's image and translation status to the README.

> (also let me know if you want to become a reviewer)

Concerning Weblate or GitHub? 
I have never done that before, but I will be very happy to do it!

MM2-0:
> I will open a PR to add Weblate's image and translation status to the README.
I was just about to do that, but if you want to, go ahead. :i18n module readme would be enough I think?

> Concerning Weblate or GitHub?
Weblate. To approve french translations. If I understand correctly, unapproved strings can be edited by anyone, but approved strings can only be edited by reviewers (other users can still make suggestions). Thing is, I won't approve strings in a language I don't speak so I think it would be best to have at least one reviewer in each supported language (though I think it's not strictly necessary)

PR #185, opened 2022-10-15
Add permission ACCESS_RESTRICTED_SETTINGS

daywalk3r666:
This is needed on android13 for device- and app-notification permission. This is block for sideloaded apps


jamal2362:
Bunp

PR #240, opened 2023-01-29
Fix icon pack detection

daywalk3r666:
This fix this issue https://github.com/LawnchairLauncher/lawnicons/issues/989


Sir-Photch:
Why did you close this?

daywalk3r666:
Because it isn't working. Etar was only themed because i had my own themed icon app(kvaesitso icons) installed

PR #241, opened 2023-01-29
Add layout-option to restrict rotation to portrait mode

Sir-Photch:
If I understand correctly, this would block the "rotate-button" to pop up every time I am lying on my side.

As suggested in: https://stackoverflow.com/questions/2730855/prevent-screen-rotation-on-android

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/SharedLauncherActivity.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/SharedLauncherActivity.kt
@@ -1,6 +1,7 @@
 package de.mm20.launcher2.ui.launcher
 
 import android.app.WallpaperManager
+import android.content.pm.ActivityInfo
 import android.content.res.Configuration
 import android.content.res.Resources
 import android.os.Bundle
@@ -35,6 +36,7 @@ import androidx.compose.ui.unit.dp
 import androidx.core.view.WindowCompat
 import androidx.core.view.WindowInsetsControllerCompat
 import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.Observer
 import androidx.lifecycle.flowWithLifecycle
 import com.google.accompanist.systemuicontroller.rememberSystemUiController
 import de.mm20.launcher2.globalactions.GlobalActionsService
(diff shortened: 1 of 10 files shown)


MM2-0:
It's possible that there are users that use the launcher in landscape mode (e.g. tablet users). You'd break it for them if you added `android:screenOrientation="portrait"`

Also, wrong activity, you'd need to add this to https://github.com/MM2-0/Kvaesitso/blob/main/app/ui/src/main/AndroidManifest.xml#L7-L27

Sir-Photch:
Well, tablet screens are even more rectangular in shape. That's why I think it makes even less sense to use it in landscape mode in that case.

The rotation button is just annoying imo. I did not see that on any other launchers.

jamal2362:
It would be better to do it through the code so that the user can decide what he wants to have.

A button in the settings where you can activate that the screen no longer rotates.

Sir-Photch:
Well adding a switch in the settings is more complex than just modifying a manifest. I'd be willing to try this though, if anyone could point me to the right source files 😅

core/preferences/src/main/proto/settings.proto
```diff
@@ -298,6 +298,7 @@ message Settings {
     bool bottom_search_bar = 2;
     bool reverse_search_results = 3;
     bool fixed_search_bar = 4;
+    bool fixed_rotation = 5;
```
Sir-Photch, line 301: So I had a look on how I could add this programmatically in the layout settings. I've found that a lot of settings-methods are generated; And looking in the commit-log it seemed to me that I had to add this line for these methods to be generated. This somehow is not working though. What am I missing?

MM2-0: They are generated at build time, either run the project or press the "build project" button in Android Studio should fix it.

Sir-Photch: Ah, I had to remove my calls to those non-existing methods and build it for it to generate, thanks 👍🏼

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/SharedLauncherActivity.kt
```diff
@@ -132,6 +134,15 @@ abstract class SharedLauncherActivity(
                             val reverseSearchResults by viewModel.reverseSearchResults.observeAsState(false)
                             val fixedSearchBar by viewModel.fixedSearchBar.observeAsState(false)
 
+                            viewModel.fixedRotation.observe(this) { fixedRotation ->
+                                requestedOrientation = if (fixedRotation) {
+                                    ActivityInfo.SCREEN_ORIENTATION_PORTRAIT + ActivityInfo.SCREEN_ORIENTATION_NOSENSOR
```
Sir-Photch, line 139: Assuming here that `SCREEN_ORIENTATION`-values work as flags

MM2-0: I think so but please use bitwise or (`or`) to combine the flags

Sir-Photch: Well, that did not work anyway. :)

Sir-Photch:
I've tested it on my Pixel 7 Pro and it seems to be working as I intended. Is there anything else that needs to be added?

MM2-0:
Looks good to me. Preference migration would be required if the default value was `true` but in that case, it's not necessary because an unset boolean defaults to `false` anyway.

Issue #171, opened 2022-09-16
Favorites disappear below the clock, when the low battery / battery charging indicator shows up.


MSBack:
This behaviour seems to be a bug or at least unwanted, in my case certainly. It would be better if the battery indicator simply inserted itself between the favorites bar and the clock, thereby raising the clock a bit higher.

To be clear, I don't have the favorites widget enabled, but I have enabled the setting to show favorites under the clock ("show the first row of pinned items"). "Fill screen height" is also enabled.

MSBack:
Great, appreciate the responsiveness and look forward to seeing it implemented. And yeah, in the mean time, I've deactivated the low / charging battery indicator.

PR #246, opened 2023-02-04
Add additional row in vertical and horizontal pager in horizontal clock mode showing favorites

Sir-Photch:
As suggested in #171. I figured people that use the `Compact`-preset, well, value compactness; That's why I came up with a new layout instead of modifying the existing ones. Currently, it looks like this when selected:

<details>
<summary>Screenshot</summary>

2706da2a-4460-41fd-98c1-1828a72539bc

</details>

I re-used the `DynamicZone`; In the bottom case it's just having access to a `FavoritesPartProvider`. Settings whether to show date/favorites are blocked when `Extended` is selected. `Horizontal`/`Vertical` remain unchanged. 

There is probably a more descriptive name than "Extended" out there, though. 😃
(Maybe "Stacked"?)


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/ClockWidget.kt
```diff
@@ -74,10 +75,22 @@ fun ClockWidget(
                     }
 
                     DynamicZone(
-                        modifier = Modifier.padding(bottom = 16.dp),
+                        modifier = if (layout == ClockWidgetLayout.Extended) {
+                            Modifier.padding(top = 16.dp)
+                        } else {
+                            Modifier.padding(bottom = 16.dp)
+                        },
```
Sir-Photch, line 82: I am not sure about these paddings, but they seem to look fine.

MM2-0:
I like how it looks, I assume it shows the date when you stop the music playback? My idea for the compact layout was to use a horizontal pager. Swipe left/right to switch between favorites and the dynamic part. That way we don't need a third layout option.

Sir-Photch:
> I assume it shows the date when you stop the music playback?

Well, the upper `DynamicZone` still has access to the original `partProvider`, so yes:

<details>
<summary>Screenshot</summary>

372fe596-d147-48c1-9922-78f64f816cc6

</details>

> That way we don't need a third layout option.

Well, maybe there are users that like the option to only have one dynamic part? Currently, this is just another option to choose from.

MM2-0:
> Well, maybe there are users that like the option to only have one dynamic part? Currently, this is just another option to choose from.

In that case they can simply disable the favorites part. I think it was a mistake to make the favorites part just another dynamic part. It disrupts the workflow when the favorites are sometimes there and sometimes not. If you enable the favorites part, you expect them to always be there. But let me make a poll, let's see what other users think.

Sir-Photch:
In any case, what about merging this and implementing the horizontal-pager in another PR?

Sir-Photch:
I re-implemented it, ditching the third clock-layout in favor of favorites always being there if the option is checked in settings. Also, I have added the horizontal-pager to the compact layout. However, now the date is centered and the DynamicZone for the favorites has a different height than the DynamicZone for everything else. See my attached screenshots.

<details>
<summary>Screenshot comparison</summary>

Collage Maker-07-Feb-2023-10 31-PM

</details>

How can I enforce the same height across both DynamicZones?

MM2-0 (commented):
Looks good, I think we can merge this. Thanks for your help.

Sir-Photch:
So your changes fixed the height problem in the compact layout?

Also, the `datePart` just stays in the settings-protobuf, even if it is not settable anymore?

MM2-0:
> So your changes fixed the height problem in the compact layout?

Yes, should be fixed.

> Also, the `datePart` just stays in the settings-protobuf, even if it is not settings UI anymore?

Ah, you're right, let me remove it.

PR #250, opened 2023-02-09
Add LazyRow in weather-widget displaying hourly forecasts

Sir-Photch:
I had another idea how Kvaesitso's UX could receive some more love, namely the weather widget:

<details>
<summary>Screenshot</summary>

0d7c0c7d-8349-454d-bcb9-f4116ff986c8

</details>

I removed the time-selector and instead added a `LazyRow` that enables users to horizontally scroll through a (hopefully soon settable) number of hourly forecasts. Also, I rearranged the remaining parts a little bit, the "main forecast" displayed in the top part is now the daily average, with min/max temperatures.

Let me know what you think about this idea. I'd say that this is probably not ready to merge though, since I changed more than in my previous PRs and me being new to Kotlin / Android means that I probably butchered some typical practices. :baby:

Questions that remain:

- [ ] Do something about the date-selector / "details" expandable.
- [ ] Come up with a better alternative for getting the conditions for a given day, instead of selecting the conditions for the day as the conditions at noon
- [ ] Maybe convert it to a three-column layout such that the dailyAverage-icon is centered properly
- [ ] Reason about adding a setting that specifies how many hours should be displayed in the lazyrow
- [ ] Maybe add a scrollbar at the bottom indicating the current position


data/weather/build.gradle.kts
```diff
@@ -50,5 +50,6 @@ dependencies {
     implementation(project(":core:preferences"))
     implementation(project(":core:permissions"))
     implementation(project(":core:i18n"))
+    implementation(project(mapOf("path" to ":core:base")))
```
MM2-0, line 53: I've never seen that, is there any difference between this and the lines above?

core/base/src/main/java/de/mm20/launcher2/helper/IterableUtils.kt
```diff
@@ -0,0 +1,10 @@
+package de.mm20.launcher2.helper
+
+object IterableUtils {
+    inline fun <T, R : Comparable<R>> List<T>.medianBy(crossinline selector: (T) -> R?): T? {
+        if (this.isEmpty())
+            return null
+
+        return this.sortedBy { selector(it) } [ this.size / 2 ]
+    }
+}
```
MM2-0, line 10: 1. This should be moved to a file called `Iterable.kt` in the `:core:ktx` module.
2. It doesn't need to be wrapped in an object.
3. I assume it could be generalized to `Iterable<T>.medianBy` instead of `List<T>.medianBy`

Sir-Photch:
> Not sure what to think about the top part… I think the large number should be the current temperature because that is usually the information I'm most interested in. Same goes for the large icon and the condition.

I guess this is up for debate. Usually, I just want to know what temperatures I can expect for a given day. And if I'm more interested, I have a look at the hourly forecast. This might be something that could be solved with a setting `Show daily instead of hourly forecast in weather-widget view`?

> Also, it feels a bit cluttered, I'm not a fan of the centered icon and there is a lot of whitespace in the bottom left corner.

So you prefer icon on the right and temperature in the middle? As for the whitespace: Currently, it is more or less created by the date picker. And I also think this is just space that could be used by "permanently expanded" details, if the date-picker is supposed to stay.

> As for the number of hourly forecasts: you're right, it is a good idea to limit them. Maybe only show hourly forecasts for the next 24 hours and then daily forecasts for everything after.

This would cause the date-picker to be somewhat redundant, wouldn't it?

> For reference, this is what it used to look like:

I suppose this is a previous version of Kvaesitso?

> Maybe we can find a way to replace the text labels with icons, but in a more compact way.
 
I also prefer icons over text.

MM2-0:
> I guess this is up for debate. Usually, I just want to know what temperatures I can expect for a given day. And if I'm more interested, I have a look at the hourly forecast. This might be something that could be solved with a setting Show daily instead of hourly forecast in weather-widget view?

In that case, it might be worth considering to add a second weather widget variant. Similar how it's done on iOS where you can pick one of three differently sized widgets that show a different amount of information.
make out
> So you prefer icon on the right and temperature in the middle?

I prefer a two column layout. Or at least align the items properly so that we have some kind of recognizable reading flow. But even then, three column layout could be problematic on small screens. I think it would make sense to group the condition label with the icon as they convey similar information.

> This would cause the date-picker to be somewhat redundant, wouldn't it.

Yes, pretty much. Also, I'm unsure whether we need to keep the ability to click on a forecast to load it into the main view. Maybe we should get rid of that as well or find a different way to show forecast details.

> I suppose this is a previous version of Kvaesitso?

Yes, it's a screenshot from 2018.

MM2-0:
I put together a quick draft:

<details>

Screenshot_20230210_222436

- wind icon rotates depending on the wind direction
- cloud colors match the color scheme (not sure about that)
</details>

And then, maybe we could add a switch for hiding the daily forecasts and another one to hide the hourly forecasts if you don't want the widget to take up as much space.

Sir-Photch:
That looks very nice! Which branch? Or is this just a mockup?

MM2-0:
It's a mockup.

Sir-Photch:
Closing in favor of #253.

PR #253, opened 2023-02-11
Weather-Widget: replace drop-down selectors with scrollable rows

Sir-Photch:
Had some git-headache. See #250.

Current status:

<details>
<summary>Screenshots</summary>

### Compact:
weather-widget-compact

### Non-compact:
weather-widget-expanded

</details>

I figured that the time-forecast would also have to be selectable. There is a switch in weather-widget-settings enabling users to choose which layout they would like. Otherwise, this is almost done I think? I thought about some blend-in animation when row-items scroll into view, but I have no idea what I could use to achieve this.


MM2-0:
Current status:

<details>

Screenshot_20230212-160409_Kvaesitso Debug~2

</details>

TODO:

- [ ] Check humidity, precipitation and wind for invalid values and hide icons accordingly

Sir-Photch:
Have you thought about adjusting the cloud color dynamically according to the material theme again? For some material themes, the cloud palette does stick out a bit, so I don't think this is a bad idea.

Otherwise, I had the idea to change the city icon on the top left to a ["location icon"](https://fonts.google.com/icons?selected=Material%20Symbols%20Rounded%3Amy_location%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048) if it's the latitude / longitude that is displayed.

Also, how did you generate the code for the icons? Is there some tool that you used? I can't imagine that you copied the values manually. :sweat_smile: I could add the location icon then as well, if it is not already present.

MM2-0:
> Have you thought about adjusting the cloud color dynamically according to the material theme again? For some material themes, the cloud palette does stick out a bit, so I don't think this is a bad idea.

I have tried but I wasn't very satisfied with the results. I might give it another go later.

> Also, how did you generate the code for the icons? Is there some tool that you used?

I use a script, that converts svg paths to compose:

<details>

```js
#!/usr/bin/env node

const parseSVG = require('svg-path-parser')

const d = process.argv[2]

const path = parseSVG(d)

for (const cmd of path) {
    switch (cmd.code) {
        case 'M': console.log(`moveTo(${cmd.x}f, ${cmd.y}f)`)
            break
        case 'm': console.log(`moveToRelative(${cmd.x}f, ${cmd.y}f)`)
            break
        case 'L': console.log(`lineTo(${cmd.x}f, ${cmd.y}f)`)
            break
        case 'l': console.log(`lineToRelative(${cmd.x}f, ${cmd.y}f)`)
            break
        case 'V': console.log(`verticalLineTo(${cmd.y}f)`)
            break
        case 'v': console.log(`verticalLineToRelative(${cmd.y}f)`)
            break
        case 'H': console.log(`horizontalLineTo(${cmd.x}f)`)
            break
        case 'h': console.log(`horizontalLineToRelative(${cmd.x}f)`)
            break
        case 'C': console.log(`curveTo(${cmd.x1}f, ${cmd.y1}f, ${cmd.x2}f, ${cmd.y2}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'c': console.log(`curveToRelative(${cmd.x1}f, ${cmd.y1}f, ${cmd.x2}f, ${cmd.y2}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'S': console.log(`reflectiveCurveTo(${cmd.x2}f, ${cmd.y2}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 's': console.log(`reflectiveCurveToRelative(${cmd.x2}f, ${cmd.y2}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'Q': console.log(`quadTo(${cmd.x1}f, ${cmd.y1}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'q': console.log(`quadToRelative(${cmd.x1}f, ${cmd.y1}f, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'T': console.log(`reflectiveQuadTo(${cmd.x}f, ${cmd.y}f)`)
            break
        case 't': console.log(`reflectiveQuadToRelative(${cmd.x}f, ${cmd.y}f)`)
            break
        case 'A': console.log(`arcTo(${cmd.rx}f, ${cmd.ry}f, ${cmd.xAxisRotation}f, ${cmd.largeArc}, ${cmd.sweep}, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'a': console.log(`arcToRelative(${cmd.rx}f, ${cmd.ry}f, ${cmd.xAxisRotation}f, ${cmd.largeArc}, ${cmd.sweep}, ${cmd.x}f, ${cmd.y}f)`)
            break
        case 'Z': console.log('close()')
            break
    }
}
```

</details>

> I could add the location icon then as well, if it is not already present.

A similar icon is part of the Material Icons: https://fonts.google.com/icons?selected=Material%20Icons%20Round%3Amy_location%3A so it is already present as part of the material-icons-extended library and can be used with `Icons.Rounded.MyLocation`.

Sir-Photch:
> I use a script, that converts svg paths to compose

Thanks for sharing!

> A similar icon is part of the Material Icons

Only now I see that there are two categories: Symbols and Icons. Explains why the other ones are not available.

> I have tried but I wasn't very satisfied with the results.

This also could be something for the user to unset in settings if they don't like it, couldn't it?

I've added checks for valid detail values. Also, I added a regexp for lat/long matching.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/weather/WeatherWidget.kt
```diff
@@ -430,6 +443,25 @@ fun WeatherDaySelector(
...
+
+        if (!atLeftEnd && !atRightEnd) {
+            return 1f
+        }
+
+        val alpha = 1f - (if (atLeftEnd) viewportStartOffset - itemStart else itemEnd - viewportEndOffset) / it.size.toFloat()
+
+        if (degree != 1f) alpha.pow(degree) else alpha
+
+    } ?: 1f
```
Sir-Photch, line 463: Would this be something to put into core:ktx?

Sir-Photch:
I've added blending into / out of view:

<details>
<summary>Screenshot</summary>

blending

</details>

I think it looks pretty nice, before, items were always cut off at the beginning and end of the row. This change makes that less noticeable.
You can adjust how quickly this "blending" is applied by changing the `degree`-parameter of `LazyListLayoutInfo.blendIntoViewAlpha()`, if necessary.

The only thing still bugging me a bit is the provider-info in the top right being somewhat cut off when the card-corner-radius is high. But otherwise, I think this is good to go, isn't it?

MM2-0:
> The only thing still bugging me a bit is the provider-info in the top right being somewhat cut off when the card-corner-radius is high.

True, we should increase the padding a bit.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/weather/WeatherWidget.kt
MM2-0: I don't want Jetpack Compose dependencies in other modules so there is another `ktx` package in `de.mm20.launcher2.ui.ktx`, for extension functions for compose types.

Sir-Photch:
Another thing I've noticed:

<details>
<summary>Screenshot</summary>

Screenshot_20230214-230603

</details>

Apparently, checking  `forecast.precipitation != -1.0` is not enough to determine if there is a valid value in that field. It shows 0mm even though it is raining?

Sir-Photch:
It is the same on the latest release:

<details>
<summary>Screenshot</summary>



</details>

This seems to be an issue to OpenWeatherMap though, since for other Providers it is fine.

MM2-0:
Maybe it is less than 0.05 mm so it is rounded down to zero.

MM2-0:
Or maybe I made a mistake:

https://github.com/MM2-0/Kvaesitso/blob/ce71784b747b4169c0060dadcecd2a95dd7cc7ef/data/weather/src/main/java/de/mm20/launcher2/weather/openweathermap/OpenWeatherMapApi.kt#L90

(this should be `3h`, not `3d`).

PR #258, opened 2023-02-16
Searchbar: Launch on enter / done

### LINKED ISSUE #53: [Enhancement] Launch first app search result when tapping "Done/Search"
In order to launch an app more quickly, I would like to be able to search for it and then just tap `→`. 
If no apps match the search, it could launch the first app shortcut found.

I've used this a lot in other launchers and kinda miss it.

Sir-Photch:
Closes #53.

Currently, it will only launch the first app that is found. I suppose it would make sense to add other searchables as well.

Things I'd still want to add:
- [x] Pass search query to first quickaction, if no other searchable is found


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -69,7 +73,23 @@ class SearchVM : ViewModel(), KoinComponent {
...
+    fun launchBestMatch(context: Context) {
+
+        /*
+        if (launchOnEnter.value != true)
+            return
+        */
+
+        suspend {
+            searchJob?.join()
+        }
```
MM2-0, line 85: What do you expect this to do?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -65,11 +73,40 @@ class SearchVM : ViewModel(), KoinComponent {
...
+        if (!launchOnEnter)
+            return
+
+        val launched = listOf(
+            appResults,
+            appShortcutResults,
+            contactResults,
+            calendarResults,
+            fileResults
+        ).map { it.value ?: emptyList() }
```
MM2-0, line 97: This can be optimized using `.firstNotNullOfOrNull { it.value?.firstOrNull() }`

MM2-0:
Looks good so far. One thing we should add is some sort of visual indication to let users know beforehand which item will be launched. But I can add that later.

Sir-Photch:
> One thing we should add is some sort of visual indication to let users know beforehand which item will be launched. But I can add that later.

Yes, I have also thought about this: [iOS is doing something similar](https://support.apple.com/en-us/HT201285) where they put a border around the best match if it is an app. I think something like this would look nice as well. For the case when no match is found, you could default to highlighting the first quickaction.

Sir-Photch:
I've added the back-end necessary for highlighting, and a rough prototype how I imagine it to look like for the AppResults-GridView:

<details>
<summary>Screenshot</summary>

searchbar

My idea will obviously only work for non-square icon shapes right now. Also, the circle is cut off.
Though judging by your changes for #253, I am sure you have better Ideas how to integrate this into the launchers appearance :)

</details>

Things missing:

- [ ] Highlighting for `ListItem`
- [ ] Default highlighting of first quick-action

Sir-Photch:
For squared icons with rounded corners it looks nice, but for square icons it doesn't fit as well:

<details>
<summary>Screenshot</summary>

highlight

</details>

Maybe it'd be possible to choose this outline-box based on what the current icon-shape is?

MM2-0:
> Maybe it'd be possible to choose this outline-box based on what the current icon-shape is?

It is, you can use `LocalIconShape.current` to access it. I was considering it but liked the rounded rect more. But you have a good point, and with that in mind, adapting the selection box to the icon shape is objectively the better solution.

<details>

photo_2023-02-18_18-51-42

photo_2023-02-18_18-51-43

photo_2023-02-18_18-51-44 (2)

photo_2023-02-18_18-51-44


</details>

Sir-Photch:
I've changed it to the icon shape. One weird thing I have noticed with the reuleaux-triangle:

<details>
<summary>Screenshot</summary>

reuleaux

</details>

Otherwise, highlighting looks fine.
Also, I have adjusted the border-size to `Dp.Hairline`, since otherwise I find it too pronounced.

PR #267, opened 2023-02-23
Search: reorder results by launchCount

### LINKED ISSUE #223: Smart (frequency+recency) and instant search everything
Thanks a lot for this very nice and complete launcher!

I like very much the ability to search a bit everything (have all _on_ except Wikipedia and websites preview), but struggle to understand the underlying mechanisms and think it could be improved.

1. When I insert only 1 letter, I don't get app suggestions, contacts or shortcuts, only folders or files (if _on_), which is a pity and not what most people want, I think.
2. I need to type sometimes up to 3 characters (sometimes 2 are enough) to get the app I want and even more for a file.
3. Maybe the most important: my choices are not remembered, although you seem to have this mechanism implemented for the favorites.

Am I right? If yes, do you have any ideas on how to implement this? Although it doesn’t search files, the [KISS launcher](https://github.com/Neamar/KISS/) have this implemented for its app/shortcuts/contacts search and its source is open, so I guess inspiration could be taken there. Sesame shortcuts, which is more complete but unfortunately closed source, have this for all, so it means it should be possible.

Sir-Photch:
Feels like the logical consequence of #258 and closes #223.

I think there is some overhead we could eliminate by reducing the limit, e.g. `.getRanksByLaunchCount(limit = 50)`. I am not quite clear though if this would impact reordering negatively? We would have to make sure that there is at least one result for each `SavableSearchable.domain`.


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -165,34 +170,47 @@ class SearchVM : ViewModel(), KoinComponent {
...
-                                contacts,
-                                wikipedia,
-                                website,
-                                files,
-                                actions
-                            ).firstNotNullOfOrNull { it.firstOrNull() }
-                        }
+                        ranksByLaunchCount.collectLatest {
+
+                            for ((domain, ranks) in it.groupBy { it.domain }) {
```
Sir-Photch, line 175: Do you think it would be a performance improvement if we already ordered the results at the database-level? E.g. `GROUP BY 'type'`? Or is kotlin already fast enough?

Sir-Photch:
Added a switch in settings for users to decide. This should be done now.

data/favorites/src/main/java/de/mm20/launcher2/favorites/FavoritesRepository.kt
```diff
@@ -286,6 +291,18 @@ internal class FavoritesRepositoryImpl(
         }
     }
 
+    override fun getRanksByLaunchCount(limit: Int): Flow<List<Pair<Int, SavableSearchable>>> {
```
MM2-0, line 294: It's unnecessary to query the entire `SavableSearchable` object. Query the keys only to reduce the overhead.
Also don't use `Pair` in public APIs, it makes it difficult to understand, what exactly is returned here. Use a new, more descriptive data class as return type.

MM2-0: Nevermind, it's used. Let's see if we can optimize that away.

Sir-Photch: No you are right, it should be enough to just return something like a new dataclass just containing `launchCount` and `key`. After all, reordering just happens by keys, that are ordered by their launch count

MM2-0: But we could use the `key`, `type`, and `launchCount` fields from the `SavedSearchableEntity` and put those in a new data class, so we don't need to deserialize the actual `Searchable`

Sir-Photch: This new class wouldn't be an entity though, right? Does it still make sense to put it into `database.entities`?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -165,34 +173,64 @@ class SearchVM : ViewModel(), KoinComponent {
...
-                                files,
-                                actions
-                            ).firstNotNullOfOrNull { it.firstOrNull() }
-                        }
+                        ranksByLaunchCount.collectLatest {
+
+                            if (query.isNotEmpty() && reorderByRelevance.value) {
+                                val fileRanks = mutableListOf<Pair<Int, SavableSearchable>>()
+
+                                for ((domain, ranks) in it.groupBy { it.second.domain }) {
```
MM2-0, line 181: grouping could be performed before sharing the flow  (L80), e.g. 

```kt
.map { it.groupBy { it.second.domain } }`
.shareIn(viewModelScope, ...)

```

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -72,6 +75,10 @@ class SearchVM : ViewModel(), KoinComponent {
         .getHiddenItemKeys()
         .shareIn(viewModelScope, SharingStarted.Eagerly, replay = 1)
 
+    private val ranksByLaunchCount = favoritesRepository
+        .getRanksByLaunchCount()
+        .shareIn(viewModelScope, SharingStarted.Eagerly, replay = 1)
```
MM2-0, line 80: Should be shared lazily, we don't need to query data when ordering by relevance is not enabled.

app/ui/src/main/java/de/mm20/launcher2/ui/settings/search/SearchSettingsScreen.kt
```diff
@@ -194,21 +200,40 @@ fun SearchSettingsScreen() {
...
+        }
+        item {
+            PreferenceCategory {
                 Preference(
                     title = stringResource(R.string.preference_screen_tags),
                     summary = stringResource(R.string.preference_screen_tags_summary),
+                    icon = Icons.Rounded.Tag,
                     onClick = {
                         navController?.navigate("settings/search/tags")
                     }
```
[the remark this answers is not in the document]
MM2-0: Looks good but imho hidden items should be grouped with tags.

MM2-0:
I have changed the sorting algorithm:
- rankings are determined for all items, not just the top items
- pinned status is included in ranking, pinned items are ranked higher than non-pinned.

Sir-Photch:
> pinned status is included in ranking, pinned items are ranked higher than non-pinned.

Are you sure about that? I thought that this is what the favorites-widget (or the favorites that show up when nothing is searched) is for.

Otherwise, looks good to me :thumbsup: 
I had other ideas for different "sorting modes" anyway. But I wanted to open an issue for that.

MM2-0:
> Are you sure about that?

Well I thought people would mark their apps as favorites because they are relevant for them. Then again, if they will also have a higher launch count so maybe it is not necessary.

Sir-Photch:
> Well I thought people would mark their apps as favorites because they are relevant for them.

Of course, but again, I think that this is somewhat already represented by the favorites row, if enabled. 

> What alternative sorting modes do you have in mind?

This launchCount approach (or at least, my initial one) has the obvious problem that new apps won't ever be considered if other apps are just, well, overrepresented. Your idea with `pinned` also being considered kind of mitigates this (but not anymore actually, since `launchCount` is the priority now anyway. Maybe `pinned > launchCount` would actually make more sense...)

#223 mentions some kind of "recent"-mode, where the usage of X last days / weeks is only taken into account. But since we don't track timestamps of launches in the DB, we would have to add a new table, with entities having two values for example:

```sql
create table LaunchTimestamps(
    key String,
    timeStamp dateTime primary key
)
```
(not sure about the datatypes, we could also just use some kind of unix time if we really want to nerd out I guess)
`time` should suffice as primary key? This way, it would be pretty easy to get launchcounts in a given timespan. This timespan could even be user-configurable, with some kind of slider that goes from 1 week in 1-week increments to maybe, idk, 2 months?

The only other thing that would be necessary is to discard old entries. I wouldn't know where to put something like this, like a "database cleanup job". I suppose just purging old elements that are older than the specified limit would be wasteful, if the user increases this threshold. But otherwise, I think this table would get big really quickly?

MM2-0:
> But otherwise, I think this table would get big really quickly?

Probably, that's why I've opted against such a table.

What I'd like to do is to change the way that launchCount is calculated, to account for recent use, not for total use over the entire time.

Sir-Photch:
Probably not, you can consider any alpha $\in\left(0,1\right)$ and rewrite the first expression as:

$$
c_{i+1} = c_i - \alpha c_i + \alpha
$$

This will converge to 1.
And the second one to

$$
c_{i+1} = c_i - \alpha c_i
$$

which will converge to 0.

MM2-0:
$c$ would start at $0$, so $c_n \in 0..1$ as well.

>This will converge to 1.
And the second one to

Which is good, isn't it? This way, newly installed apps have a chance to catch up and no app can fall too far behind.

> Would there be one with grouping / couting a 5kB database? Genuine question. This would be the exact solution, that's why I'm interested. (I thought that databases were made for this?)

You're right. Probably not. We're talking about a few hundred entries here. Should be fine. Let's not over optimize things that aren't problematic.

Sir-Photch:
> Which is good, isn't it? This way, newly installed apps have a chance to catch up and no app can fall too far behind.

Ah yes, I thought the start would be the current launchCount.
If we start from 0 anyway, we could keep the current launchCount and introduce a new column, something like `weight`. And based on that, we could add a new "weighted"-ordering.

But still, I like my database idea; I think I will play around with this and submit a PR if it does not turn out to be rubbish. 

This way, with the additional weighted sorting, we would have four ways of ordering the search results:

1. Alphabetic, "oldschool"
2. Total count, most persistent (current "relevance")
3. Recent count, somewhat persistent, but up to date, based on set time threshold
4. Weight, very dynamic, based on the $\alpha$ we (or users?) choose

Which should be enough for anyone complaining that KISS has better search ordering 😜 

Anyway, I would not consider pinned items in this ordering. In my usecase, my pinned apps don't necessarily have to be the ones that I use most often, but more like the ones I want myself to use more. (Like Signal/Molly > WhatsApp) So I don't think this fits here. But again, I'm just one user of many.

In the future, we could even add a widget that displays some stats about app usage. But this is probably a long way off :D

MM2-0:
> Ah yes, I thought the start would be the current launchCount.

We'd need some kind of migration.

One thing we haven't considered yet is that Android itself keeps usage stats for installed apps: https://developer.android.com/reference/android/app/usage/UsageStatsManager

But I'm not sure if third party apps can use this. One one hand, [PACKAGE_USAGE_STATS](https://developer.android.com/reference/android/Manifest.permission#PACKAGE_USAGE_STATS) is declared as `signature|privileged|development|appop|retailDemo` (whatever `development|appop|retailDemo` is). On the other hand, I have several apps installed which request that permission (Home Assistant, Lawnchair, Automate). Might be worth to look into it.

Sir-Photch:
> One thing we haven't considered yet is that Android itself keeps usage stats for installed apps

But only for apps, right?

There is one function that is probably interesting: 
[`UsageStatsManager.queryAndAggregateUsageStats(startTime, endTime)`](https://developer.android.com/reference/android/app/usage/UsageStatsManager#queryAndAggregateUsageStats(long,%20long)),
that returns a map with keys as package names and values being usage statistics, which in turn have the method 
[`UsageStats.getTotalTimeVisible()`](https://developer.android.com/reference/android/app/usage/UsageStats#getTotalTimeVisible()).

We could also order by that. But considering that this needs deep android permissions, I would add this "android-app-statictics" feature as an optional setting for the "recent count / total count" option.

> But I'm not sure if third party apps can use this.

If Lawnchair can, I am sure there is a way to integrate this.

PR #272, opened 2023-03-01
Search: introduce weighted reordering

Sir-Photch:
As discussed in #267.

Things missing:

- [ ] ~Coming up with a constant, `maxTimespanMs` (1 Month?) and put it somewhere (where?) such that the database can access it and cleanup entries older than `currentTimeMs - maxTimespanMs`~
- [ ] ~Investigate [`UsageStatsManager.queryAndAggregateUsageStats(startTime, endTime)`](https://developer.android.com/reference/android/app/usage/UsageStatsManager#queryAndAggregateUsageStats(long,%20long))~
- [x] New settings-screen with advanced options for search
- [ ] (actually run it and see if it works)


MM2-0:
I've given this a second thought and honestly, I don't really see a benefit in having both, a "weight" and a full record of recent launches. Both are attempts to solve the same problem, which is identifying which items are currently most used. "Sort by last used" as a third sort mode could be a useful addition but for that, it is enough to store the last launch timestamp per item, no full history is needed.

And let's scrap that idea of using UsageStatsManager, you are right, it only refers to apps so it isn't that useful for us.

core/preferences/src/main/proto/settings.proto
```diff
@@ -225,9 +225,13 @@ message Settings {
...
     enum SearchResultOrdering {
       Alphabetic = 0;
-      Relevance = 1;
+      TotalLaunchCount = 1;
+      RecentLaunchCount = 2;
+      Weighted = 3;
     }
     SearchResultOrdering search_result_ordering = 5;
+    int64 recent_launch_count_timespan_ms = 6;
+    double weight_factor = 7;
```
MM2-0, line 234: There is no real benefit for storing this in user settings. It only makes it harder for us, in case we want to tweak that value in a future update.

Sir-Photch: I intended these ones to be user-configurable. I was referring to an upper limit for `recent_launch_count_timespan_ms`, which would be used by the database for cleanup.

Like: `maxTimespanMs = 1 Month`, so the database would always keep 1 month of timestamps. And `recentLaunchCountTimespanMs` could be any value between one day and `maxTimespanMs`, which would be the value used for reordering.

The same goes for the weight-factor, if users wanted a faster change they could increase alpha, otherwise, decrease it.

Sir-Photch:
> "Sort by last used" as a third sort mode could be a useful addition but for that, it is enough to store the last launch timestamp per item, no full history is needed.

I suggest ditching the timestamp idea in favor of the weights then, since that one would keep configurability with a settable alpha. This "sort by recent use" option only makes sense for me if users could define how recent they want their results to be.

MM2-0:
> This "sort by recent use" option only makes sense for me if users could define how recent they want their results to be.

What do you mean? Just sort all items by their "last use" time (maybe cap it at 30 days) and sort all other items after.

> To be more concrete, we could implement a slider with a stepsize of sth. like 0.1 for alpha between 0.1 and 0.9 (or other values that we determine), that does not necessarily say something like "change decay alpha" but more like "faster change" and "slower change".

That's not the kind of parameter that you want to expose to users. It's the developer's job to find a good value. Giving users a slider is just a lame excuse to move that responsibility to the user. In reality, it takes days until one can even see an effect after changing that value. No one dedicates that much time to tweak such a parameter. Maybe it makes sense for development purposes (we could hide it in the debug menu) but end users shouldn't be bothered to deal with that.

Sir-Photch:
> > This "sort by recent use" option only makes sense for me if users could define how recent they want their results to be.
> 
> What do you mean? Just sort all items by their "last use" time (maybe cap it at 30 days) and sort all other items after.

I am talking about the approach with a settable alpha / timespan threshold. But whatever, lets focus on the other argument:
 
> That's not the kind of parameter that you want to expose to users. It's the developer's job to find a good value. Giving users a slider is just a lame excuse to move that responsibility to the user. No one dedicates that much time to tweak such a parameter.

I agree and disagree. My take would be that our job would be to find bounds for this alpha and, following your argumentation, a good _default_ value. I don't think it is a bad idea to expose this to users. Someone sideloading an app from github / adding another repository to their fdroid client is more likely to be a tinkerer that wants to change stuff, don't you think? I doubt that users that "just don't care" would even think or know about the option to change their default launcher. 

An option that users could switch between "slow", "default" and "fast", with preset values determined by us would be a good compromise.

MM2-0:
> An option that users could switch between "slow", "default" and "fast", with preset alpha values determined by us would be a good compromise.

That's something I could agree with. But even then, I wouldn't store a decimal value in user settings but an enum instead, and lookup the actual values at runtime. And these settings should be moved out of `SearchBarSettings`. `search_result_ordering` as well, I think.

core/database/src/main/java/de/mm20/launcher2/database/migrations/Migration_21_22.kt
```diff
@@ -0,0 +1,13 @@
...
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+class Migration_21_22: Migration(21, 22) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("""
+            ALTER TABLE `Searchable`
+            ADD `weight` DOUBLE NOT NULL DEFAULT 0.0
+            """)
+    }
+}
```
Sir-Photch, line 13: Is this migration fine? I had an issue where the launcher crashed with an exception related to migration:

<details>
<summary>Stacktrace</summary>

```
java.lang.IllegalStateException: Migration didn't properly handle: Searchable(de.mm20.launcher2.database.entities.SavedSearchableEntity).
                                                                                                     Expected:
                                                                                                    TableInfo{name='Searchable', columns={launchCount=Column{name='launchCount', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, pinned=Column{name='pinned', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, hidden=Column{name='hidden', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, weight=Column{name='weight', type='REAL', affinity='4', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, type=Column{name='type', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, key=Column{name='key', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=1, defaultValue='undefined'}, searchable=Column{name='searchable', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}}, foreignKeys=[], indices=[]}
                                                                                                     Found:
                                                                                                    TableInfo{name='Searchable', columns={key=Column{name='key', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=1, defaultValue='undefined'}, type=Column{name='type', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, searchable=Column{name='searchable', type='TEXT', affinity='2', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, launchCount=Column{name='launchCount', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, pinned=Column{name='pinned', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, hidden=Column{name='hidden', type='INTEGER', affinity='3', notNull=true, primaryKeyPosition=0, defaultValue='undefined'}, weight=Column{name='weight', type='DOUBLE', affinity='4', notNull=false, primaryKeyPosition=0, defaultValue='0.0'}}, foreignKeys=[], indices=[]}
                                                                                                    	at androidx.room.RoomOpenHelper.onUpgrade(RoomOpenHelper.kt:94)
                                                                                                    	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.onUpgrade(FrameworkSQLiteOpenHelper.kt:252)
                                                                                                    	at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:416)
                                                                                                    	at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:316)
                                                                                                    	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getWritableOrReadableDatabase(FrameworkSQLiteOpenHelper.kt:231)
                                                                                                    	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.innerGetDatabase(FrameworkSQLiteOpenHelper.kt:189)
                                                                                                    	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getSupportDatabase(FrameworkSQLiteOpenHelper.kt:151)
                                                                                                    	at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper.getWritableDatabase(FrameworkSQLiteOpenHelper.kt:104)
                                                                                                    	at androidx.room.RoomDatabase.inTransaction(RoomDatabase.kt:629)
                                                                                                    	at androidx.room.RoomDatabase.assertNotSuspendingTransaction(RoomDatabase.kt:448)
                                                                                                    	at androidx.room.RoomDatabase.query(RoomDatabase.kt:477)
                                                                                                    	at androidx.room.util.DBUtil.query(DBUtil.kt:75)
                                                                                                    	at de.mm20.launcher2.database.SearchDao_Impl$23.call(SearchDao_Impl.java:972)
                                                                                                    	at de.mm20.launcher2.database.SearchDao_Impl$23.call(SearchDao_Impl.java:969)
                                                                                                    	at androidx.room.CoroutinesRoom$Companion$createFlow$1$1$1.invokeSuspend(CoroutinesRoom.kt:128)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
                                                                                                    	at java.lang.Thread.run(Thread.java:1012)
                                                                                                    	Suppressed: kotlinx.coroutines.DiagnosticCoroutineContextException: [StandaloneCoroutine{Cancelling}@7542527, Dispatchers.Main.immediate]
```
</details>

Uninstalling Kvaesitso and redeploying fixed this, but I think this error will reoccur.

Sir-Photch:
I've setup a small jupyter notebook to play around with different alphas:

https://gist.github.com/Sir-Photch/17b23f663e034244d9c942d602b384c0

MM2-0:
I think the weight sorting should also be used for the favorites. So what do you think about moving the factor slider to settings/search/favorites?

core/preferences/src/main/proto/settings.proto
```diff
@@ -223,11 +223,7 @@ message Settings {
     }
     SearchBarColors color = 3;
     bool launch_on_enter = 4;
-    enum SearchResultOrdering {
-      Alphabetic = 0;
-      Relevance = 1;
-    }
-    SearchResultOrdering search_result_ordering = 5;
+    reserved 5;
```
MM2-0, line 226: This was never in production (I think?) so no need to reserve this index.

core/database/src/main/java/de/mm20/launcher2/database/migrations/Migration_21_22.kt
MM2-0: I think you need to annotate the entity's field with [@ColumnInfo(defaultValue = 0.0)](https://developer.android.com/reference/androidx/room/ColumnInfo.html#defaultValue()) in order to make the schemas match.

core/preferences/src/main/proto/settings.proto
Sir-Photch: this was merged after the last release so yeah probably

Sir-Photch:
> I think the weight sorting should also be used for the favorites. So what do you think about moving the factor slider to settings/search/favorites?

Probably makes sense since weight is adjusted independently of search reordering.

core/database/src/main/java/de/mm20/launcher2/database/migrations/Migration_21_22.kt
Sir-Photch: I've checked by deploying current main and then installing this branch's version, and this seems to be fine now. Maybe you also check to make sure?

MM2-0: Seems to be working

core/preferences/src/main/java/de/mm20/launcher2/preferences/migrations/Migration_12_13.kt
```diff
@@ -10,5 +11,10 @@ class Migration_12_13: VersionedMigration(12, 13) {
                     .setDatePart(true)
                     .build()
             )
+            .setResultOrdering(
```
MM2-0, line 14: Default value should also be added to `Defaults.kt` since migration only run for existing users.

Sir-Photch:
This should be good to go, is anything missing?

MM2-0:
We still need a migration to populate the `weight` column with existing data from the `launchCount` column so that users don't lose their frequently used items when they update. I think the easiest would be `launchCount / max(launchCount)` which would ensure that the order is kept.

Also, I'm still not a 100% satisfied with the labeling of the preferences but I haven't found something better yet either.

MM2-0:
> Favorites are still ordered after launchCount though?

Oh true, that is also still missing

Sir-Photch:
What do you mean?
Currently, it's 
```sql
ORDER BY pinned DESC, weight DESC, launchCount DESC
```

MM2-0:
Ah, sorry, now I understand your message. Yes, you are right, it is sorted by both weight and launchCount, so the sorting doesn't immediately change after the update. But it's extremely unstable and as soon as you launch *any* item, it will immediately move to the top of the list. I think items that are currently at the top should get an advantage.

Sir-Photch:
Okay so I adjusted the migration and gave it a couple of spins in the emulator with some breakpoints, it seems to be fine.

I tried different approaches, most notably:
```sql
UPDATE Searchable SET weight = launchCount / (SELECT MAX(launchCount) FROM Searchable);
```
but I couldn't get this to work properly, so I split the queries.

PR #276, opened 2023-03-07
Make `OrbitClock` somewhat respect orbital dynamics

Sir-Photch:
This might be nitpicking but I had some time on my hands and thought about reordering the "planets" in the orbit-clock to cure my OCD. Also, I added a seconds-"planet", since I think it is nice to have some small speedy planet spicing up your home screen.

$T_{hours} > T_{minutes} \Leftrightarrow r_{hours} > r_{minutes}$  holds, that's why I swapped hours and minutes. 
I tried [Kepler's third law](https://en.wikipedia.org/wiki/Kepler%27s_laws_of_planetary_motion#Third_law) but the ratios wouldn't look good. 🤡

grafik

[Video](https://streamable.com/4pow40)

Another small thing I have noticed: since the font is rendered transparent, it is hard to read on backgrounds that have brighter areas. But I couldn't figure out how to solve this, since the fonts already have `alpha = 1f`. Also, I've not added the seconds-planet to the horizontal layout since I figure it'd be to cluttered.


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/clocks/OrbitClockVM.kt
```diff
@@ -0,0 +1,21 @@
+package de.mm20.launcher2.ui.launcher.widgets.clock.clocks
+
+import androidx.compose.runtime.mutableStateOf
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import org.koin.core.component.KoinComponent
+
+class OrbitClockVM : ViewModel(), KoinComponent {
```
MM2-0, line 10: This is terribly inefficient. `viewModelScope` is only cancelled when the ViewModel is cleared, meaning this will continue to update even when the launcher is not in foreground. You should move this to a `LaunchedEffect` inside the composable and wrap it in `lifecycle.repeatOnLifecycle(...) { ... }` to make sure that this is paused and restarted when the launcher is paused / resumed.

MM2-0:
Looks good but I'm not a hundred percent sure about the radii. I feel like the minutes circle should be a bit smaller to put the three circles in golden ratio. Maybe make the hour orbit a bit larger and the minute orbit a bit smaller and resize the "planets" accordingly so that the space between each of them is equal (at 12:00:00)

<details>

223536559-96f4afe4-4072-4b93-9e23-be5e55020c67

</details>

> Another small thing I have noticed: since the font is rendered transparent, it is hard to read on backgrounds that have brighter areas.

True, but
- You can probably still make out the rough position of the "planets"
- If the wallpaper is mainly bright, the clock turns dark. You can also enable that manually if you are having trouble to read the clock on the wallpaper.
- You have a second clock in the status bar.

There is also an issue with the preview in clock widget settings where the text cuts through the dialog surface but whatever. They have added blend mode support for text [recently](https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.4.0-beta02), idk if that helps to fix that.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/clocks/OrbitClockVM.kt
Sir-Photch: I've changed it, but I am not sure if that is what you meant. I found similar logic in ProvideCurrentTime.kt.

Still, how can we make sure the coroutine within the Orbit-Clock's `repeatOnLifecycle` is canceled when a different clock-layout is chosen? The code in ProvideCurrentTime could be changed to update every second, but I think this is somewhat wasteful for clock layouts that don't need second accuracy.

MM2-0: `LaunchedEffect` (and its children) are canceled when it leaves the composition.

MM2-0: > The code in ProvideCurrentTime could be changed to update every second, but I think this is somewhat wasteful for clock layouts that don't need second accuracy.

Another approach is used for [dynamic clock icons](https://github.com/MM2-0/Kvaesitso/blob/0adef99d4e65293af1f558d150fc59aa69a5333e/app/ui/src/main/java/de/mm20/launcher2/ui/component/ShapedLauncherIcon.kt#L334):

There, the seconds hand is moved by a 60s animation which is restarted whenever the time (provided by LocalTime) is updated.

Sir-Photch:
> to put the three circles in golden ratio.

This is how it looks like if the radii are in the golden ratio:

$r_{i+1} = \varphi \cdot r_i$

Bildschirmfoto vom 2023-03-08 13-49-30

I think what you are asking for is that the distance between the radii are in the golden ratio?

```kotlin
val rs = size.width * 0.1f
val rm = size.width * 0.24f
val rh = rm + (rm - rs) * PHI_F
```

grafik

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/clocks/OrbitClockVM.kt
Sir-Photch: > There, the seconds hand is moved by a 60s animation which is restarted whenever the time (provided by LocalTime) is updated.

I've thought about introducing another orbit layout like this too, with smooth animations and orbit paths similar to these [typical loading circles](https://dm0qx8t0i9gc9.cloudfront.net/thumbnails/video/IevwMJs/loading-circle-gradient-w-alpha-25fps_s6zya32g_thumbnail-1080_01.png ), for them to look like comet-trails. Wasn't sure about this though, if it would fit into the launcher's design so I figured to change the current one first and then think about it again. For this, the smooth animation would be pretty handy, yes. What do you think?

Sir-Photch:
Spacing between planets is equal now and the difference in radii are a golden ratio.
I've also managed to fix the opacity-issue.

This is how it looks now:

grafik

You can adjust `rs, rm, sSize, mSize` as you like and it will keep the ratios for `rh, hSize`.

Sir-Photch:
I've changed it to smooth animations: [Video](https://streamable.com/wnqe0f)

This is how the spacing looks when the "planets are aligned":

Bildschirmfoto vom 2023-03-09 22-33-56

Sir-Photch:
Is anything particular keeping this from being merged @MM2-0?

MM2-0:
Not really, just didn't find the time to review the latest changes yet.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/clocks/EmptyClock.kt
MM2-0, line 1: Wait, why did you delete this?

MM2-0: Ah I get it. It's not really necessary. I would leave the `when` branch though, so that it doesn't look like we forgot it.

MM2-0:
Oops, I think rebasing this on main wasn't a smart move by me lol.

PR #422, opened 2023-06-08
add Oneplus clock widget

eliavandini:
I understand that there are many OnePlus users who may not appreciate the red color option available in the OnePlus clock widget. However, after having it on my home and lock screen for years, I have developed a fondness for it. Nevertheless I genuinely enjoy using this launcher and appreciate the digital clock widgets it already offers. However, I do miss having a touch of red on my home screen. That's why I have made the decision to create a pull request that introduces a new digital clock widget. In this widget, the hour digits will be colored red specifically when they represent the number one.
Screenshot_2023-06-08-11-57-35-43_b50c6ccf92c546ed20b65090255147e8
Screenshot_2023-06-08-11-57-18-88_b50c6ccf92c546ed20b65090255147e8

### DIFF
--- a/core/preferences/src/main/proto/settings.proto
+++ b/core/preferences/src/main/proto/settings.proto
@@ -111,12 +111,12 @@ message Settings {
     ClockWidgetLayout layout = 1;
     enum ClockStyle {
       DigitalClock1 = 0;
-      OneClock = 1;
-      DigitalClock2 = 2;
-      OrbitClock = 3;
-      BinaryClock = 4;
-      AnalogClock = 5;
-      EmptyClock = 6;
+      OneClock = 6;
+      DigitalClock2 = 1;
+      OrbitClock = 5;
+      BinaryClock = 2;
+      AnalogClock = 3;
+      EmptyClock = 4;
     }
     ClockStyle clock_style = 2;
     bool date_part = 3;
--- a/core/preferences/src/main/proto/settings.proto
+++ b/core/preferences/src/main/proto/settings.proto
@@ -111,11 +111,12 @@ message Settings {
     ClockWidgetLayout layout = 1;
     enum ClockStyle {
       DigitalClock1 = 0;
-      DigitalClock2 = 1;
-      OrbitClock = 5;
-      BinaryClock = 2;
-      AnalogClock = 3;
-      EmptyClock = 4;
+      OneClock = 1;
+      DigitalClock2 = 2;
+      OrbitClock = 3;
+      BinaryClock = 4;
+      AnalogClock = 5;
+      EmptyClock = 6;
     }
     ClockStyle clock_style = 2;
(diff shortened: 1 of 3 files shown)


core/preferences/src/main/proto/settings.proto
```diff
@@ -111,11 +111,12 @@ message Settings {
     ClockWidgetLayout layout = 1;
     enum ClockStyle {
       DigitalClock1 = 0;
-      DigitalClock2 = 1;
-      OrbitClock = 5;
-      BinaryClock = 2;
-      AnalogClock = 3;
-      EmptyClock = 4;
+      OneClock = 1;
```
MM2-0, line 114: Don't change the enum ordinals, it will mess up users' current settings. Just assign thr next higher number. Order matters as well because this is the order they appear in settings (which is the reason that this list is seemingly unordered)

MM2-0:
Two more thoughts:
- If I understand correctly, this only shows if the hour includes a 1. So there is a chance that a user won't notice a difference between this and the standard clock in the clock selector.
- Do you think it would make sense to use the theme's primary color instead of a hardcoded red?

garvongithub:
This could be a nice option for OnePlus users like me. Primary color option for clock as mentioned by @MM2-0 should be different and separate because the lock screen clock of OnePlus does not depend on primary color. Hour “1” of lock screen is always red, even if the primary color provided by the system is yellow or green. This clock option can provide uniformity between home screen and lock screen of OnePlus devices, paired along with system font. Please consider this.

MM2-0:
> This could be a nice option for OnePlus users like me. Primary color option for clock as mentioned by @MM2-0 should be different and separate because the lock screen clock of OnePlus does not depend on primary color. Hour “1” of lock screen is always red, even if the primary color provided by the system is yellow or green. This clock option can provide uniformity between home screen and lock screen of OnePlus devices, paired along with system font. Please consider this.

I don't mind adding this, after all it is just an extra cosmetic option that has no impact on users that don't like to use it, but I need to think of a way to communicate how this clock differs from the default clock because you only see the difference half of the day. My idea so far is that I could rewrite the clock picker so that for each clock style you have different variants. Then we could also add other variants of this clock, for example outlined, colored with primary color, OnePlus style, and so on.

garvongithub:
Yes, introducing this as a variant for Fat digital clock and Boring digital clock seems better than introducing this as a separate clock style. That way, there can also be a material you clock variant (like in pixel) for these above-mentioned styles. The design could be the like this, The text (red highlighted) could be moved above and variants (blue strokes) listing different options could be placed below it. Selecting the variant will show a preview of the variant with name or description of the variant at the top or bottom
image

MM2-0:
I have manually copied the important bits of this PR in order to account for the other new clock styles.
Closing this now.

PR #598, opened 2023-10-28
Reduce flickering of highlighted best match while searching

### LINKED ISSUE #559: Search delay when using "Launch on enter"
Currently the UI flickers a little bit when you enter a search term, since it is redrawn with every keypress. This is especially noticeable, when a search term only matches the first default quickaction. 

I suggest we add a small delay that needs to pass after a keypress, which the UI waits for and that is reset if another keypress is registered within that delay.

This delay would of course need some tuning, not to hinder UX too much in case search-results take too long to show up. It would probably integrate with haptic feedback (#532) nicely, that is triggered when search-results are updated.

How does this sound to you? Or is there a better way to do this?

Sir-Photch:
Closes #559

### DIFF
--- a/app/app/build.gradle.kts
+++ b/app/app/build.gradle.kts
@@ -13,7 +13,7 @@ android {
     }
 
     packaging {
-        resources.excludes.add("META-INF/DEPENDENCIES")
+        //resources.excludes.add("META-INF/DEPENDENCIES")
         resources.excludes.add("META-INF/LICENSE")
         resources.excludes.add("META-INF/LICENSE.txt")
         resources.excludes.add("META-INF/license.txt")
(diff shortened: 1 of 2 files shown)


app/app/build.gradle.kts
```diff
@@ -13,7 +13,7 @@ android {
     }
 
     packaging {
-        resources.excludes.add("META-INF/DEPENDENCIES")
+        //resources.excludes.add("META-INF/DEPENDENCIES")
```
MM2-0, line 16: What's this?

Sir-Photch: With this line, the app wouldn't build on my system.
I can stash this as a local change if you want this to be unchanged 🤷🏼‍♂️

MM2-0: I don't remember why I added this, probably because it wouldn't build without it, but whatever. I will uncomment it again if I need to.

PR #604, opened 2023-11-01
add workaround for resource/folder mime types when view activity fails

Sir-Photch:
This is probably specific to grapheneos, but it still fixes a long-standing bug that causes nothing to happen when a SavableSearchable result that is a directory is launched.

This happens because by default, apparently no app implements `Intent.ACTION_VIEW` for mime-type `resource/folder` on grapheneos, even though the default file browser (`com.android.documentsui`) is available.

In that case, just open `com.android.documentsui` directly with the given uri.

Also, this fixes a possible issue where the launchcount is incremented even though launching fails.

### DIFF
--- a/core/ktx/src/main/java/de/mm20/launcher2/ktx/Context.kt
+++ b/core/ktx/src/main/java/de/mm20/launcher2/ktx/Context.kt
@@ -24,6 +24,18 @@ fun Context.tryStartActivity(intent: Intent, bundle: Bundle? = null): Boolean {
         startActivity(intent, bundle)
         true
     } catch (e: ActivityNotFoundException) {
+        if (intent.type == "resource/folder") {
+            packageManager.getLaunchIntentForPackage("com.android.documentsui")?.let {
+                it.setDataAndType(intent.data, intent.type)
+                    .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_GRANT_READ_URI_PERMISSION)
+
+                return try {
+                    startActivity(it, bundle)
+                    true
+                } catch (_: Exception) { false }
+            }
+        }
+
         false
     } catch (e: SecurityException) {
         false
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
@@ -89,10 +89,11 @@ class SearchVM : ViewModel(), KoinComponent {
     fun launchBestMatchOrAction(context: Context) {
         val bestMatch = bestMatch.value
         if (bestMatch is SavableSearchable) {
-            bestMatch.launch(context, null)
-            favoritesService.reportLaunch(bestMatch)
+            if (bestMatch.launch(context, null))
+                favoritesService.reportLaunch(bestMatch)
             return
-        } else if (bestMatch is SearchAction) {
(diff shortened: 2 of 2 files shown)


Sir-Photch:
Okay, seems like `com.android.documentsui` can't open arbitrary paths. But it has it's own activity for viewing downloads; It's probably sensible to keep that?

MM2-0:
> Okay, seems like `com.android.documentsui` can't open arbitrary directories. But it has it's own activity for viewing downloads; It's probably sensible to keep that?

Yes, IIRC I have tried that before and it didn't work. Hardcoding downloads (and other paths that can be opened by intent) would make sense.

core/ktx/src/main/java/de/mm20/launcher2/ktx/Context.kt
```diff
@@ -24,6 +25,19 @@ fun Context.tryStartActivity(intent: Intent, bundle: Bundle? = null): Boolean {
...
+                .setComponent(ComponentName("com.android.documentsui", "com.android.documentsui.ViewDownloadsActivity"))
+
+            return try {
+                startActivity(aospDocsViewDownloadsIntent, bundle)
+                true
+            } catch (_: Exception) {
+                false
+            }
+        }
+
```
MM2-0, line 40: This should be moved to https://github.com/MM2-0/Kvaesitso/blob/4c0b34e091e0779a164fd60eeaa0941dcb2da916/data/files/src/main/java/de/mm20/launcher2/files/providers/LocalFile.kt#L189. Less hacky that way.

data/files/src/main/java/de/mm20/launcher2/files/providers/LocalFile.kt
```diff
@@ -184,7 +186,19 @@ internal data class LocalFile(
     }
 
     override fun launch(context: Context, options: Bundle?): Boolean {
-        return context.tryStartActivity(getLaunchIntent(context), options)
+        if (context.tryStartActivity(getLaunchIntent(context), options))
+            return true
```
MM2-0, line 190: Add curly braces or remove line break

data/files/src/main/java/de/mm20/launcher2/files/providers/LocalFile.kt
```diff
@@ -184,7 +186,19 @@ internal data class LocalFile(
...
+
+        // startsWith allows path to end with a slash
+        if (isDirectory && path == Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS).path) {
+            val aospViewDownloadsIntent = Intent()
+                .setComponent(ComponentName("com.android.documentsui", "com.android.documentsui.ViewDownloadsActivity"))
+                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_GRANT_READ_URI_PERMISSION)
+
+            return context.tryStartActivity(aospViewDownloadsIntent, options)
```
MM2-0, line 198: The package name `com.android.documentsui` might not be guaranteed. Can't you just use an implicit intent here? Action: `android.intent.action.VIEW_DOWNLOADS`, see https://cs.android.com/android/platform/superproject/main/+/main:packages/apps/DocumentsUI/AndroidManifest.xml;l=139

shtrophic: yep that works as well. thanks for the hint

PR #611, opened 2023-11-07
Add OSM search provider

### LINKED ISSUE #606: [Feature proposal] Location Search Provider
I've got an idea for another search provider: maps!

Consider this use case: I want to get the current opening-hours for my favorite café.
For this, it would be pretty handy to just enter the name which then displays a short card with information about the place, like name, address and opening hours (and, if it is open at the moment). Additionally, you could display a small static map image that shows the place on the map.

The [Overpass API](https://wiki.openstreetmap.org/wiki/Overpass_API) provides functionality to query a given string, at a given location, within a given radius.

Consider this snippet: 

```python
import requests

# 1000 meter radius around lat,lon
query = """[out:json];
node(around:1000,49.792,9.932)["name"~"pizza",i];
out;
"""

response = requests.get("https://overpass-api.de/api/interpreter", params={'data':query})

elements = response.json()['elements']
for element in elements:
    tags = element['tags']
    print(tags['name'], tags['addr:street'], tags['addr:housenumber'], tags['opening_hours'])
```

It should output something like:
```
Pizza Hut Dominikanerplatz 3b Mo-Th 11:30-21:00; Fr,Sa 11:30-22:00; PH,Su 12:00-21:00
Pizza King Würzburg Eichendorffstraße 12 Mo-Sa 11:00-14:00, 17:00-23:00; Su 11:00-23:00
Pizza Classico Am Exerzierplatz 6 Mo-Fr 11:00-14:00
Pizzamore Eichhornstraße 23 Mo-Sa 11:00-21:00; PH,Su off
Main Döner & Pizza Theaterstraße 4b Mo-Th 10:00-24:00, Fr,Sa 10:00-05:00, PH,Su 12:00-22:00
480grad neapolitan Pizza & Bar Juliuspromenade 16 Mo-Th 17:00-22:00; Fr 17:00-23:00; Sa 13:00-23:00, Su 13:00-22:00
```

Additionally, if adding a small static image would contribute to UX, there are services like [Mapbox](https://docs.mapbox.com/playground/static/) that provide an API for that. (And also, a library: https://github.com/mapbox/mapbox-java/) Things like rate-limits with one API-Key probably still have to be figured out then though.

Finally, a user quick-launching or tapping on the search result could launch an intent that opens up their preferred maps app. (Google Maps, Organic Maps, ...)

What do you think about this?

Sir-Photch:
Closes #606


Sir-Photch:
I've seen that you are working on a plugin system. Is that affecting what I am doing right now?

MM2-0:
> I've seen that you are working on a plugin system. Is that affecting what I am doing right now?

I have only superficially reviewed your changes yet, but it shouldn't affect anything. I plan to add a plugin API for every type of search result (app, shortcut, file, calendar event, contact, website, article, and, with your PR, location), but for now I'm focusing on files.
The important thing is, that we have a generic interface in `:core:base` and that the UI only ever references that interface, no specific implementations of it. That way we can later easily add other implementations for plugins. But at least from what I've seen so far, you have done that correctly.

Sir-Photch:
Okay so I'd say this is ready for comments. 

The UI is still totally botched of course, but I am sure you have some suggestions how to do that anyway. It should get my point across what I originally wanted to do:

- Display a icon specific to `LocationCategory` if set, else use a default icon
- Show the label of the location
- When details are not shown:
  - just display name 
  - address
  - distance 
  - and if the location is open.
- When details are shown:
  - say until when the location has opened (if applicable) 
  - show the next opening day with time (if applicable)
  - show distance with direction
  - (maybe also show address again)
  - show toolbar with icons to launch geo-intent and browser to website (if url is known)

Additionally, I still have some UI ideas that could use some debating:

- [ ] Add an Image in the search result when details are shown that displays a static map section, with the location in it's center? This would mean wrapping another API though. ([Mapbox](https://docs.mapbox.com/playground/static/)) And probably one with access tokens.
- [ ] Add a small tooltip when opening-hour information is displayed that shows something like "not accurate? consider changing the entry in OSM" upon tapping

Apart from the UI, the back-end should be mostly done though.
However, remaining questions I have are:

- [ ] the HTTP-client that Retrofit uses does not have timeouts set currently, since Overpass requires quite a long time for queries with high radii (which makes sense, since the area of interest increases with `radius**2`
- [ ] whether to save locations to database and add the ability to add the results to favourites. For this to work properly, the OverpassAPI would probably have to be queried independently of the users current location, just with the location ID (which is unique for every node)
- [ ] whether to filter the results further (Should we include results that have no information besides coordinates and label)
- [ ] whether to use more fallbacks when determining the label of the OsmLocation (there are many tags that could be used potentially)
- [ ] whether to sort the location search results by closest to furthest. But since the distances are calculated close to the front-end, the search logic (SearchVM) doesn't know about them. And calculating the distance so early didn't seem right to me.

Also, about the issue of adding more web-search providers; What about adding disclaimers to each of them in the settings, stating something like "Enabling this will send queries to abc.xyz"

MM2-0:
Thanks, I will review/test it tomorrow. As for your questions:

> Add an Image in the search result when details are shown that displays a static map section, with the location in it's center? This would mean wrapping another API though. ([Mapbox](https://docs.mapbox.com/playground/static/)) And probably one with access tokens.

Maybe we could grab the correct tile from the OSM tile server? (https://tile.openstreetmap.org/${z}/${x}/${y}.png). Wouldn't be centered, but maybe close enough.

>Add a small tooltip when opening-hour information is displayed that shows something like "not accurate? consider changing the entry in OSM" upon tapping

The UI shouldn't assume that the data is from OSM, so it would need to be implemented in a way that is independent from OSM.

> the HTTP-client that Retrofit uses does not have timeouts set currently, since Overpass requires quite a long time for queries with high radii (which makes sense, since the area of interest increases with radius**2

what order of magnitudes are we talking about? I think the max search radius should be chosen in a way that these timeouts don't happen. Or is that unrealistic?

> whether to save locations to database and add the ability to add the results to favourites. For this to work properly, the OverpassAPI would probably have to be queried independently of the users current location, just with the location ID (which is unique for every node)

Serializers / Deserializers shouldn't make network calls. You should store all data that you can't retrieve offline in the database (see Wikipedia and cloud file serializers). However, I see that this isn't optimal either since opening hours might change and users might rely on them to be correct. I have to think this through.

>whether to filter the results further (Should we include results that have no information besides coordinates and label)

Need to test this. Depends on how many of these results are useless.

> whether to use more fallbacks when determining the label of the OsmLocation (there are many tags that could be used potentially)

I'd say if we have the data downlonaded anyway, there is no harm in using them? Or are these alternative labels incorrect or misleading?

> whether to sort the location search results by closest to furthest. But since the distances are calculated close to the front-end, the search logic (SearchVM) doesn't know about them. And calculating the distance so early didn't seem right to me.

If we have to calculate them early in order to sort by them, I think it is fine to do so. But does the API not have any options to customize the order when making the request?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -0,0 +1,306 @@
...
+                                Icon(
+                                    modifier = Modifier
+                                        .padding(top = 8.dp, bottom = 8.dp, start = 16.dp)
+                                        .rotate(directionArrowAngle),
+                                    imageVector = Icons.Rounded.ArrowUpward,
+                                    contentDescription = null
+                                )
+                                if (distance != null) {
+                                    Text(
+                                        text = distance.metersToString(context)
```
MM2-0, line 223: There should be an option to choose between sane and imperial units somewhere.

Sir-Photch: You are right, I forgot about users of insane units :eyes:

Sir-Photch:
> Maybe we could grab the correct tile from the OSM tile server? ([https://tile.openstreetmap.org/${z}/${x}/${y}.png](https://tile.openstreetmap.org/$%7Bz%7D/$%7Bx%7D/$%7By%7D.png)). Wouldn't be centered, but maybe close enough.

That's a good idea! And I think it would also be easier to accomplish. Their [usage policy](https://operations.osmfoundation.org/policies/tiles/) specifies some requiremets though we should obey, e.g. caching some tiles locally as long as their HTTP-header specifies.  The question then is how to display an image that was rendered as 256x256 on screen, while remaining legibility. 
Also, we should then start to think about what [zoom level](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels) to use. There are also [some examples](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Implementations) how to convert `lat,lon,zoom` to tile indices.


> The UI shouldn't assume that the data is from OSM, so it would need to be implemented in a way that is independent from OSM.

Well, I guess that's object orientation. An easy "hack" would be to define some method like `getErrorReportUrl()` in the `Location` interface. For OSM, it would be https://www.openstreetmap.org/fixthemap.

I really think that this is something we should add, especially when then kvaesitso users will start using the location functionality, as a way to give back to the OSM project.


> what order of magnitudes are we talking about? I think the max search radius should be chosen in a way that these timeouts don't happen. Or is that unrealistic?

AFAIK there are some default values set when constructing a retrofit instance, so they are enough. I think the latency will depend on the current user location, since every node in the radius is matched against the query. Currently in Würzburg, I get around 3-4 seconds for radius values higher than 1.5km, which seems to be the sweetspot.

You can try how it works at your location.


> Serializers / Deserializers shouldn't make network calls. You should store all data that you can't retrieve offline in the database (see Wikipedia and cloud file serializers). However, I see that this isn't optimal either since opening hours might change and users might rely on them to be correct. I have to think this through.

For OSM, they return a unique ID that can later be used to query the exact node again, as I understood. If we need to cache the tiles, we should then keep track of what node introduced a new tile anyway (or something alike).

> I'd say if we have the data downlonaded anyway, there is no harm in using them? Or are these alternative labels incorrect or misleading?

It seems like there are no concrete standards for osm data, so some values present under some tag might not necessarily be trying to describe the same thing on a different node under the same tag. Missing information like "name" is unusual though, especially for shops and restaurants.

> If we have to calculate them early in order to sort by them, I think it is fine to do so. But does the API not have any options to customize the order when making the request?

No, I couldn't find how to do this with OverpassQL.

MM2-0:
> Also, about the issue of adding more web-search providers; What about adding disclaimers to each of them in the settings, stating something like "Enabling this will send queries to abc.xyz"

Yes, with a link to OSM's privacy policy.

> The question then is how to display an image that was rendered as 256x256 on screen, while remaining legibility of labels

Oh true, I didn't consider that. We might need 6 images and stitch them together.

> AFAIK there are some default values set when constructing a retrofit instance, and currently they seem fine. The latency will probably depend on the current user location, since every node in the radius is matched against the query. Currently in Würzburg, I get around 3-4 seconds for radius values higher than 1.5km, which seems to be the sweetspot.

Do we actually need a timeout? Or can we just cancel the previous request once another one comes in? I think this is what I did in Wikipedia search.

> Well, I guess that's object orientation. An easy "hack" would be to define some method like getErrorReportUrl() in the Location interface. For OSM, it would be https://www.openstreetmap.org/fixthemap.

> I really think that this is something we should add, especially when then kvaesitso users will start using the location functionality, as a way to give back to the OSM project.

Okay, agreed.

> For OSM, they return a unique ID that can later be used to query the exact node again, as I understood. If we need to cache the tiles, we should then keep track of what node introduced a new tile anyway (or something alike).

What for? Shouldn't it be enough to store the URL of the tile and when it expires? And just rely on [Coil's built-in cache mechanism](https://coil-kt.github.io/coil/image_loaders/#caching).

MM2-0:
I tested it now, looks good so far. Some things I noticed:

- the default value for search radius was 0 after I upgraded to this debug build. Seems like a preference migration is missing
- search is pretty slow for 5 km radius, but with 1.5 km it's fine. I'd say we make 1.5 km the default value, and maybe we should add a notice that a greater search radius leads to slower search results
- i think would be better if it would show the complete opening hours, not just the next time it opens
- do you have any idea what this is:
Screenshot_20231110-130750_Kvaesitso Debug
The numbers seem to be counting down. This only occured for one search result, others were fine. Maybe because it opens in less than an hour?

I will start working on a design for the UI now.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -0,0 +1,306 @@
...
+    val userLocation by viewModel.userLocation.collectAsState(null)
+    val distance = userLocation?.distanceTo(location.toAndroidLocation())
+
+    DisposableEffect(location) {
+        viewModel.init(location, iconSize.toInt())
+        viewModel.startLocationUpdates(context)
+
+        onDispose {
+            viewModel.stopLocationUpdates(context)
+        }
```
MM2-0, line 82: There should be a lifecycleOwner.repeatOnLifecycle somewhere here to ensure that location updates are stopped when the user navigates away from the launcher (or turns off the screen)

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -0,0 +1,306 @@
...
+            }
+            AnimatedVisibility(showDetails) {
+                DisposableEffect(showDetails) {
+                    if (showDetails)
+                        viewModel.startHeadingUpdates(context)
+
+                    onDispose {
+                        viewModel.stopHeadingUpdates(context)
+                    }
+                }
```
MM2-0, line 156: Same here, this should react to Android lifecycle

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/common/SearchableItemVM.kt
```diff
@@ -170,4 +182,114 @@ class SearchableItemVM : ListItemViewModel(), KoinComponent {
...
+        trueNorthHeading.value = null
+    }
+
+    fun stopLocationUpdates(context: Context) {
+        context.getSystemService<LocationManager>()?.removeUpdates(locationListener)
+
+        // also applies to this
+        userLocation.value = null
+        geomagneticField = null
+    }
```
MM2-0, line 294: This works, but the more elegant solution (imho) would be to only expose two methods `fun getUserLocation(context: Context): Flow<Location?>`/`getUserHeading(context: Context): Flow<Float?>` and then use a [`callbackFlow`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/callback-flow.html) to manage subscription to the location service. Then you could simply use [`Flow.collectAsStateWithLifecycle`](https://developer.android.com/reference/kotlin/androidx/lifecycle/compose/package-summary#(kotlinx.coroutines.flow.StateFlow).collectAsStateWithLifecycle(androidx.lifecycle.Lifecycle,androidx.lifecycle.Lifecycle.State,kotlin.coroutines.CoroutineContext)). This would also eliminate the need for a `repeatOnLifecycle` (see my comments in `LocationItem.kt`).

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -0,0 +1,306 @@
...
+
+private fun LocationCategory?.getImageVector(): ImageVector = when (this) {
+    LocationCategory.RESTAURANT -> Icons.Rounded.Restaurant
+    LocationCategory.FAST_FOOD -> Icons.Rounded.Fastfood
+    LocationCategory.BAR -> Icons.Rounded.LocalBar
+    LocationCategory.CAFE -> Icons.Rounded.LocalCafe
+    LocationCategory.HOTEL -> Icons.Rounded.Hotel
+    LocationCategory.SUPERMARKET -> Icons.Rounded.LocalGroceryStore
+    else -> Icons.Rounded.Place
+}
```
MM2-0, line 306: This should be moved to `Location::getPlaceholderIcon`.
This would allow
  1. us to use some color for the icons / use different colors for different categories
  2. icons to respect the user's themed icon preference
  3. users to customize the icons

Icons need to be provided as XML Vector Drawables because we don't have access to Compose icons at this level. See `File.kt` for an example.

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
```diff
@@ -0,0 +1,215 @@
...
+                // convert index from sublist context
+                nextGroupIndex += nextTimeIndex
+
+                parseGroup(blocks.subList(0, nextGroupIndex))
+                blocks = blocks.subList(nextGroupIndex, blocks.size)
+            }
+
+            return openingTimes
+        }
+    }
```
MM2-0, line 201: I would move this to the OSM module and make it an `internal fun`.

core/i18n/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:locale="en">
+    <string name="location_open">Open</string>
+    <string name="location_closed">Closed</string>
+    <string name="location_open_until">Open until %1$s</string>
+    <string name="location_open_next">Opens in %1$s</string>
+    <string name="location_open_next_day">Opens next %1$s</string>
```
MM2-0, line 7: Add comments for translators to explain what the placeholders stand for.

core/preferences/src/main/java/de/mm20/launcher2/preferences/DataStore.kt
```diff
@@ -44,5 +44,6 @@ internal fun getMigrations(context: Context): List<DataMigration<Settings>> {
         Migration_15_16(),
         Migration_16_17(),
         Migration_17_18(),
+        Migration_18_19(),
```
MM2-0, line 47: You need to bump `SchemaVersion` or this migration won't run.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmLocation.kt
```diff
@@ -0,0 +1,78 @@
...
+            Intent(
+                Intent.ACTION_VIEW,
+                Uri.parse("geo:$latitude,$longitude?q=${Uri.encode(label)}")
+            ),
+            options
+        )
+    }
+
+    override fun getPlaceholderIcon(context: Context): Nothing {
+        TODO()
```
MM2-0, line 47: TODO

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
```diff
@@ -0,0 +1,21 @@
+package de.mm20.launcher2.openstreetmaps
+
+import de.mm20.launcher2.search.SavableSearchable
+import de.mm20.launcher2.search.SearchableDeserializer
+import de.mm20.launcher2.search.SearchableSerializer
+
+class OsmLocationSerializer : SearchableSerializer {
+    override fun serialize(searchable: SavableSearchable): String {
+        searchable as OsmLocation
+        TODO()
```
MM2-0, line 10: TODO

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    override val typePrefix: String
+        get() = "osmlocation"
+}
+
+class OsmLocationDeserializer : SearchableDeserializer {
+    override suspend fun deserialize(serialized: String): SavableSearchable? {
+        TODO()
```
MM2-0, line 19: TODO

data/wikipedia/build.gradle.kts
```diff
@@ -15,6 +15,7 @@ android {
 
     buildTypes {
         release {
+            isMinifyEnabled = false
```
MM2-0, line 18: Why?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
Sir-Photch: I thought that `DisposableEffect` handles this situation by calling `onDispose` when the effect leaves composition. Is that not the case when the screen is turned off? (when the user navigates away?)

data/wikipedia/build.gradle.kts
Sir-Photch: I blame android studio.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
MM2-0: No, pausing the activity does not destroy the composition. `onDispose` is called when you remove it from the composition but that doesn't happen when you pause the activity.

Sir-Photch:
> i think would be better if it would show the complete opening hours, not just the next time it opens

Sure, if things don't get too crowded.

> The numbers seem to be counting down. This only occured for one search result, others were fine. Maybe because it opens in less than an hour?

If the location is closed but opens later on the current day, a "countdown" until opening up is displayed.
The seconds should be left out of course, and the sign should be changed also.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/common/SearchableItemVM.kt
Sir-Photch: In case of using a callback flow, do I still need to specify `lifecycleOwner` or does it default to the correct one?

MM2-0: - You use a `callbackFlow` to make sure that the listener is correctly installed when the flow is started and removed when the flow is stopped
- You use `collectAsStateWithLifecycle`. This automatically starts the flow when the given lifecycle state (resumed by default) is reached and terminates it when the lifecycle falls below that state.
- When the flow is terminated, the `awaitClose` lambda is called. Here you unregister the listener
- Once the activity is resumed again, the flow is restarted and the code inside `callbackFlow` runs again

`collectAsStateWithLifecycle` uses the `LocalLifecycleOwner` by default which is the current activity, so no need to specify it further.

Use it like this in your composition:

```kt

val userLocation by remember(context) { viewModel.getUserLocation(context) }.collectAsStateWithLifecycle()

```

It's important to `remember` the result of `getUserLocation` because it is a function call if you didn't remember it, you would create a new flow object on every recomposition.

Sir-Photch: Okay, because I just had `val userHeading by viewModel.getUserHeading(context).collectAsStateWithLifecycle(null)` and it wouldn't work anymore. I appreciate the guidance.

Sir-Photch: Also, would you advise against using `trySendBlocking`? I always have the idea that blocking calls in UI-code causes issues, but nonblocking calls may need a buffer for unread values

Sir-Photch: Whatever, the nonblocking variant seems to work fine without any buffers.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
Sir-Photch: relatively simple, just save the unique ID of the location

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
Sir-Photch: I implemented this with network calls now. Otherwise, we may also just store ID, lat and lon and deserialize that (but information would be incomplete).

Anyway, we need to query OSM sooner or later anyway when retrieving deserialized data, since as you mentioned, we might display outdated information otherwise. The question is where we implement this call.

MM2-0: Bad idea. The launcher relies on serializers completing within milliseconds because this potentially runs very frequently (every time anything changes in the Searchable database table, so every time an item is pinned, unpinned, hidden, or launcher). If it doesn't complete in time, the entire launcher is slowed down by this.

Here is an alternative approach that might work:
- store ID, name, lat, lon, and category in the database
- add `suspend fun`s to the Location interface to access information that is subject to change (i.e. `suspend fun getOpeningHours`)
- load the information lazily the first time it is requested by the UI, then cache them for the lifetime of the Location object

Sir-Photch:
I've added some "live-sorting via distance" by implementing some callbacks. Do you think this is a good idea? You can test this rather easily in the emulator by changing the position and observing the results reordering themselves.

For this we might have to reason about how to order the location results, if the best match (determined deeper in the backend by SearchVM) is also a location result. I've implemented it such that this best match gets shown first anyway, even if there is a result that has a lower distance.

core/preferences/src/main/java/de/mm20/launcher2/preferences/Defaults.kt
```diff
@@ -191,6 +191,13 @@ fun createFactorySettings(context: Context): Settings {
             Settings.AnimationSettings.newBuilder()
                 .setCharging(true)
         )
+        .setLocationsSearch(
+            Settings.LocationsSearchSettings.newBuilder()
+                .setEnabled(false)
+                .setUseInsaneUnits(false)
```
MM2-0, line 197: Use R.bool.default_imperialUnits here

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -107,11 +98,164 @@ fun LocationItem(
...
+                        leftActions = listOf(
+                            DefaultToolbarAction(
+                                label = stringResource(id = R.string.menu_back),
+                                icon = Icons.Rounded.ArrowBack
+                            ) {
+                                onBack()
+                            }
+                        ),
+                        rightActions = listOfNotNull(
+                            websiteUrl.value.runCatching {
```
MM2-0, line 232: What are you catching here? `Uri.parse`? I think that only ever throws when the provided string is `null` (in which case a simple `?.let` would do the trick)

MM2-0, line 232: What are you catching here? `Uri.parse`? I think that only ever throws when the provided string is `null` (in which case a simple `?.let` would do the trick)

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -102,7 +105,7 @@ class SearchVM : ViewModel(), KoinComponent {
     }
 
     private var searchJob: Job? = null
-    fun search(query: String, forceRestart: Boolean = false) {
+    fun search(query: String, context: Context?, forceRestart: Boolean = false) {
```
MM2-0, line 108: I don't like how you change my function signatures. Especially that context being nullable here screams ugly hack to me. Here is a better approach:

- Add a new class `UserLocationProvider`, inject context as a constructor parameter
- Add a Koin declaration for that class `single<UserLocationProvider>{ UserLocationProvider(androidContext()) }` where you provide the application context as a dependency
- Inject an instance of that class into `SearchVM` using Koin's `by inject()`

Sir-Photch: So the context-instance doesn't change during runtime? I'll change that. 
Anyway, if there are any other changes that you don't like, tell me, no hard feelings. Many of these commits follow the "make it work, then make it right"-mentality anyway :)

MM2-0: I usually try to avoid passing contexts around whenever possible. There's two reasons for this:

1. Contexts are an easy way to create memory leaks. You always need to make sure that you don't store activity context references in places that outlive that activity (or the activity can't be garbage collected when it is destroyed). That's the reason we don't store a direct reference to the context in our ViewModel. The exception is the application context, which lives as long as the application process is alive. This is what Koin injects when we call `androidContext`. Technically, there is `AndroidViewModel` which provides an application context in its constructor, but I consider this an antipattern two, because
2. Contexts are way too platform specific. Yes, we develop an Android app, but we should still try to avoid using them in public interfaces. Because
   1. It's more difficult to write unit tests if the method depends on a context (joke's on me who has written like two unit tests in total for this repo. I will add them eventually, I promise!)
   2. Once you start to require them, you will always be confronted with the question "from where do I get a context now?". And then you infect your entire code with it, because, well, the obvious solution is to make the user of that function require a context as well. We can see that in your code, you found yourself asking the exact same question and your solution was to make the context nullable.

So essentially my philosophy is that I don't use `Context`s in the public interface of a class[^1], except
- if we explicitely need an activity context (e.g. `SavableSearchable.launch`, need the context we launch from)
- in classes that are essentially data classes (e.g. `SavableSearchable.loadIcon`) – I don't want to store context references in data classes so I need to provide an external one.

In both cases, the context reference must be released after the function returns.

In all other cases, I keep the public interface context-free and keep it hidden as an implementation detail that is injected by Koin if needed.

[^1] I  say public interface because technically the cleanest solution to your problem would be to write an `interface UserLocationProvider` and a separate implementation `class UserLocationProviderImpl: UserLocationProvider`, but that's overkill if you only have one implementation. And if you ever find yourself in need of a second implementation it is pretty easy anyway to separate the interface from the implementation.

Sir-Photch: And the difference between `search()` and e.g. `requestCalendarPermission()` is that `search` creates a `Job` that might capture that context and this way the reference may outlive the activity, while `requestCalendarPermission` doesn't?

I've now changed it to a new module in `core`, thanks for the thorough explanation. Still, wouldn't it make sense for `SearchVM` to implement some disposable interface that then cancels the `Job`? (Java doesn't know `IDisposable` like C#, but kotlin isn't java, amirite?)

> interface UserLocationProvider and a separate implementation class UserLocationProviderImpl: UserLocationProvider

I've stuck with a class with an internal constructor now.

> joke's on me who has written like two unit tests in total for this repo. I will add them eventually, I promise!

Testing in production also works I suppose 🤠

MM2-0: > And the difference between search() and e.g. requestCalendarPermission() is that search creates a Job that might capture that context and this way the reference may outlive the activity, while requestCalendarPermission doesn't?

The reason `requestCalendarPermission` has a context parameter is that you need an activity context to launch permission requests. I would prefer it if I could call PermissionsManager.requestPermission without a context parameter, but that doesn't work in Android.

`search` on the other hand isn't tied to any activity and we can easily use the application context every time we need one. And since Koin knows that application context, it can inject it and we don't need to worry about it.

But yes, `search` potentially outliving the activity is definitely another reason why we should avoid it here.

Sir-Photch:
Backend should be done

Sir-Photch:
Also, I came across this issue using the vertical pager layout:

https://github.com/MM2-0/Kvaesitso/assets/47949835/b3e2bbac-447d-45c9-91f4-8c8c84aa3bd2

the search-bar doesn't seem to "stick" to the top of the search results properly.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchColumn.kt
```diff
@@ -108,6 +115,9 @@ fun SearchColumn(
     val favoritesEditButton by favoritesVM.showEditButton.collectAsState(false)
     val favoritesTagsExpanded by favoritesVM.tagsExpanded.collectAsState(false)
 
+    // is it correct to use remember here?
+    val locationPriorities = remember { mutableStateMapOf<String, Int>() }
```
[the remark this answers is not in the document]
MM2-0: But I'm not a fan of this sorting mechanism. Honestly, I would just scrap it. I doubt that the average user navigates to a place using only this search result screen. Keep in mind that search results are cleared after 30 seconds of inactivity, if the screen is turned off or the user navigated away.

app/ui/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="preference_search_locations_show_position_on_map_summary">Shows your location in the map preview</string>
+</resources>
```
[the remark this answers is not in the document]
Sir-Photch: I believe I already moved this? Were you reviewing the latest commits?

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
```diff
@@ -0,0 +1,216 @@
...
+import java.time.LocalDate
+import java.time.LocalTime
+
+interface Location : SavableSearchable {
+
+    val latitude: Double
+    val longitude: Double
+    val fixMeUrl: String?
+
+    suspend fun getCategory(): LocationCategory?
```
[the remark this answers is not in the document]
Sir-Photch: Hm. I get the motivation that this would reduce some overhead when just getting the icon, but at the same time, OpenStreetMaps can be edited (and therefore: corrected) by anyone. So actually, I am not a fan of storing anything else than the location ID in the database, since everything may change. I guess we can just always store the category, but then update it as other methods like `getStreet()` are called?

core/i18n/src/main/res/values/strings.xml
```diff
@@ -855,4 +871,22 @@
     <string name="shortcut_label_unavailable">Unavailable</string>
     <!-- %1$s: app name -->
     <string name="shortcut_unavailable_description">This shortcut is unavailable because %1$s isn\'t the default launcher</string>
+    <string name="length_unit">Unit of length</string>
+    <string name="imperial">Imperial</string>
+    <string name="metric">Metric</string>
+    <string name="preference_search_location_custom_overpass_url">Overpass Url</string>
+    <string name="overpass_url">https://overpass-api.de</string>
```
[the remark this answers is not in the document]
Sir-Photch: > Is this language dependant? Is there any language where you would use a different default URL?

https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances

There are some other instances that provide the API as well. Most notably, there is an instance that only provides data for switzerland, so in this case it is "language" dependent? (More like, region dependent)

Especially concerning the tile-server there are multiple instances that provide locale-specific images, like `openstreetmap.de` and `.fr`.

> And is that URL even used? As far as I see it, you only use it in the UI, but the actual URL is still hardcoded.

You mean as the default value in the settings migration? I wasn't sure if I could use the resources there as well.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmRepository.kt
```diff
@@ -0,0 +1,166 @@
...
+        hasLocationPermission.collectLatest locationPermission@{ locationPermission ->
+            if (!locationPermission) return@locationPermission
+
+            withContext(Dispatchers.IO) {
+                httpClient.dispatcher.cancelAll()
+            }
+
+            val userLocation = poseProvider.getLocation().firstOrNull() ?: return@locationPermission
+
+            dataStore.data.map { it.locationsSearch }
```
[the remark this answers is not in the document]
Sir-Photch: Not sure how to use `combine()` here and make it function correctly. I just reordered the logic.

Sir-Photch:
What's your status on this? Do you need some input from me?

MM2-0:
> What's your status on this? Do you need some input from me?

I think I'll take it from here. I will get back to you if I have any questions.

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
```diff
@@ -133,29 +133,21 @@ abstract class Location : SavableSearchable {
...
 
     fun distanceTo(otherLocation: Location): Float =
         this.distanceTo(otherLocation.toAndroidLocation())
-
-    override fun compareTo(other: SavableSearchable): Int {
-        if (other !is Location || other.lastDistance == null || this.lastDistance == null)
-            return super.compareTo(other)
-
-        return lastDistance!!.compareTo(other.lastDistance!!)
-    }
```
Sir-Photch: I suggest keeping this logic. With this override, the secondary criterion for sorting of search results is the distance; Which makes sense, since the pizza store 100m closer might be more relevant than "ABC-Pizza" that is further away. (Assuming that otherwise, entries are sorted by name alphabetically.) This ties nicely into the comparisons in `SearchVM` that defaults to `a.compareTo(b)` if neither of the searchables have a relevance.

I see that this won't work in interfaces. So maybe implement this in `OsmLocation`? I just put this into Location since distances are nothing specific to OSM, but to locations in general. (Again, assuming we would ever have another location provider :b)

Sir-Photch: I sort of implemented a best-effort approach, by populating `lastDistance` every time `distanceTo(other)` was called:

`Location.kt`:
```kotlin
fun distanceTo(androidLocation: android.location.Location): Float {
    lastDistance = androidLocation.distanceTo(this.toAndroidLocation())
    return lastDistance!!
}
```

Now some months later, I see that this method should rather be called something like `distanceToUser()` since the logic behind it assumes that the passed location is the user's location. (This was probably implicitly given, since the parameter is of type `AndroidLocation`. But then, there is also `distanceTo(Location)`...)

Also, my approach increases that _spicy_ non-determinism. I judged `compareTo()` to be non-critical though for anything else than sorting.

Since you also seem to be in favor of caching, this seemed like the best approach to me. We compute this distance when the user is initially searching for something that is not in the database, in `OsmRepository.kt`, 140ff:

```kotlin
result
    .filter {
        !settings.hideUncategorized || (it.category != null && it.category != LocationCategory.OTHER)
    }
    .groupBy {
        it.label.lowercase()
    }
    .flatMap { (_, duplicates) ->
        // deduplicate results with same labels, if
        // - same category
        // - distance is less than 100m
        if (duplicates.size < 2) duplicates
        else {
            val luckyFirst = duplicates.first()
            duplicates
                .drop(1)
                .filter {
                    it.category != luckyFirst.category ||
                        it.distanceTo(luckyFirst) > 100.0
                } + luckyFirst
        }
    }.sortedBy {
        it.distanceTo(userLocation)
    }.toImmutableList()
```

I see that there needs to be some other approach though when we receive something that is de-serialized by the database.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
```diff
@@ -37,7 +46,9 @@ class OsmLocationDeserializer : SearchableDeserializer {
             openingSchedule = null,
             websiteUrl = null,
             phoneNumber = null,
-            isCacheUpToDate = false,
+            updatedSelf = deferred {
+                osmRepository.get(id).firstOrNull()
+            }
```
Sir-Photch: It is probably a good idea to remove the entry from the database if `.get(id)` returns `null`, since entries in OSM could also just as well be deleted. (I am not sure if IDs are re-used, but still.)

Just imagine a shop closing; That's enough of a reason for someone to delete that node in OSM.

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
Sir-Photch: Maybe we could ask `DevicePoseProvider` for some cached location, and otherwise just don't sort them via distance? Users that have location-search / weather enabled are accessing their location all the time anyway. And in case there is no location available, we loose some UX, but nothing major breaks. 

I suppose for this to work correctly, weather-providers should also access user-pose via `DevicePoseProvider` in order to have this centralized piece of code, right? (But let's not dive into refactoring-hell amirite)

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
MM2-0: It could also mean that the user is offline, or that there was an error in the request

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
MM2-0: > My approach increases that spicy non-determinism I suppose. I judged compareTo() to be non-critical though for anything else than sorting. But I see that there needs to be some other approach when we receive something that is de-serialized by the database.

Actually, this could be an issue. The [`Comparable`](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html) interface requires the relationship to be transitive (so that $` a < b \land b < c \implies a < c `$). This is not the case when the results of `compareTo` are not stable. When we update `lastDistance` while another thread is sorting a list of locations, the order would be inconsistent and I don't know how the Android Runtime would handle such a situation.

> Maybe we could ask DevicePoseProvider for some cached location, and otherwise just don't sort them via distance? Users that have location-search / weather enabled are accessing their location all the time anyway. And in case there is no location available, we loose some UX, but nothing major breaks.

This might be the best solution. Give the location an "assumed user location", which is constant for the lifetime of the `Location` object. Sort by that value. Realistically, the lifetime of a `Location` object isn't that long – unless it's pinned to favorites but in that case it isn't sorted by distance.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
Sir-Photch: I mean in case everything else works and the api returns 404. IIRC this is the behaviour for IDs that are not found.

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
Sir-Photch: Are only pinned/launched items saved to the launcher database? Because if not, we might have some Locations that have no sorting information, which will sort of "penalize" them for not coming from the API.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
Sir-Photch: Something like this:

```kotlin
fun requestUpdatedDeferredSearchable() {
    ...
    viewModelScope.launch {
        this@SearchableItemVM.searchable.value = updated.await().fold(
            onSuccess = { v -> v },
            onFailure = { e -> if (e is HttpException) null else return@launch }
        )
    }
}
```

core/base/src/main/java/de/mm20/launcher2/search/Location.kt
Sir-Photch: > But in that case we can simply aquire the last known location from DevicePoseProvider and populate the restored location with that, right?

You mean during deserialization? SGTM!

Even then I'd argue we do not have a hard requirement on locations being the same, 1m difference doesn't matter. However, if for some reason the deserialization is quicker than the API query we might end up with a location for deserialized instances that is much further away than some new location returned for the API query. This does sound like an edge case though...

MM2-0: > However, if for some reason the deserialization is quicker than the API query we might end up with a location for deserialized instances that is much further away than some new location returned for the API query. This does sound like an edge case though

It's pretty likely that that happens. The edge case here is that someone actually assigns a tag or custom label to a location… I would say if a location with a custom label matches the query, we just show it, no matter how far away it is. If someone goes through the steps of assigning a custom label, and then searches for a term that matches that label label, I'd imagine that they *want* to find that result. And even if not, it's just a bit more noise, it's not like it would hide any results from the API because of that.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
MM2-0: > What about changing updatedSelf in `DeferredSearchable<T>` from `Deferred<T?>?` to `Deferred<Result<T>>?`?

Sounds good. But maybe we use a custom class instead. We need to differentiate between three different outcomes:
1. The data is retrieved successfully. The UI is updated with the new data. The new data should be fed back into the database
2. The data is temporarily unavailable (for example due to the user being offline). The UI should continue to show the old data, possibly with a hint that the result is cached (?)
3. The data is permanently unavailable. The data should be removed from the database.

> ```kt
>onFailure = { e -> if (e is HttpException) null else return@launch }
 >```

`HttpException` is an implementation detail specific to Retrofit. Not all providers use Retrofit internally (also I consider migrating to Ktor in the future, but that's a different story). Either we implement our own exception classes (`PermanentlyUnavailableException` and `TemporarilyUnavailableException`), or we use a different `sealed interface Result` with `Result.Success` `Result.PermanentlyUnavailable` and `Result.TemporarilyUnavailable`

Sir-Photch: Okay, seems reasonable. But currently, how would we make sure that `SearchableItemVM` "unrenders" itself? `searchable` is of type `MutableStateFlow<SavableSearchable?>`, so we could assign `null` to its value. As far as I can see though every code accessing `value` defaults to some initial value passed, e.g. in `ListItem()`:
```kotlin
fun ListItem(item: SavableSearchable)
    ...
    val item = viewModel.searchable.collectAsState().value ?: item
```
(Maybe just change that to `:? return`?)

If this is unnecessary I guess we could just call `favoritesService.searchableRepository.delete(invalidLocation)` in `requestUpdatedDeferredSearchable()` and call it a day.

Anyway; If you are pressed on time I can come up with an implementation for this. My experience just tells me that two people working on the same branch can cause some headaches... 😄

MM2-0: > Okay, seems reasonable. But currently, how would we make sure that SearchableItemVM "unrenders" itself?

They are stored in a cache with a max size of 500 and should be garbage collected when they aren't needed anymore.

> As far as I can see though every code accessing value defaults to some initial value passed, e.g. in `ListItem():`

Yeah, not sure about that. Basically, I want the searchable to come from the viewmodel so that we can update it from there, but the viewmodel is only populated in a `LaunchedEffect` in `ListItem`. Since side effects do not run synchronously, there is a chance that `viewModel.searchable` is `null` for a frame. So in that frame, we use the original parameter instead. There is probably a cleaner way to do this but I can't think of one right now.

> If this is unnecessary I guess we could just call favoritesService.searchableRepository.delete(invalidLocation) in requestUpdatedDeferredSearchable() and call it a day.

`searchableRepository` shouldn't be public, I don't know why it is, probably an oversight. But we could expose a new function that forwards that call to delete, yes. And another one that updates its contents

> Anyway; If you are pressed on time I can come up with an implementation for this. My experience just tells me that two people working on the same branch can cause some headaches... 😄

Sure, go ahead. There is plenty of other things I need to attend to before we can finally release this update.

Sir-Photch:
Okay so I incorporated the changes that I suggested. 
I also added a subscript icon to the location result view when it is in its expanded view that is displayed on non-success. What do you think?

Since now we discriminate between `UpdateResult.{Success, TemporarilyUnavailable, PermanentlyUnavailable}` it is possible to retry on `TemporarilyUnavailable` -- as the name implies. There is one issue though: `Deferred<T>` does not facilitate retrying; At least I could not find any API enabling me to do so. This way, re-opening a pinned item will not actually execute the deferred block passed to the `DeferredSearchable` again, so we are stuck with the `TemporarilyUnavailable` from its first execution. Do you have a solution for this, from the top of your head?

What about just changing `updatedSelf` from `Deferred<UpdateResult<T>>?` to `(suspend () -> UpdateResult<T>)?`? `updatedSelf.await()` is called in a CouroutineScope right now anyway, isn't it?

Whether we keep the out-of-date icon or not doesn't change the fact that retrying on a network connection that was offline but might have come online in the meantime is probably the right thing to do.

Maybe, it would also make sense to show some popup to the user in case we encounter `PermanentlyUnavailable`? Otherwise, I could imagine some issues being opened up here on github saying "muy fav place iz gone!".

Anyway, If you want to make changes yourself again, go ahead.

MM2-0:
> Since now we discriminate between UpdateResult.{Success, TemporarilyUnavailable, PermanentlyUnavailable} it is possible to retry on TemporarilyUnavailable -- as the name implies. There is one issue though: Deferred<T> does not facilitate retrying; At least I could not find any API enabling me to do so. This way, re-opening a pinned item will not actually execute the deferred block passed to the DeferredSearchable again, so we are stuck with the TemporarilyUnavailable from its first execution. Do you have a solution for this, from the top of your head?

True. Also, we might want to refresh some items again at a later point. After all, searchables can live pretty long (several days), if they are in favorites.

> What about changing updatedSelf from Deferred<UpdateResult<T>>? to just (suspend () -> UpdateResult<T>)?? updatedSelf.await() is called in a CouroutineScope right now anyway, isn't it?

In that case I would make it a regular member function of the interface, not a property. And there's the question of at what level we implement the caching. Do we just leave it to each implementation and let it decide by its own whether and for how long to cache updated data?

> Maybe, it would also make sense to show some popup to the user in case we encounter PermanentlyUnavailable? Otherwise, I could imagine some issues being opened up here on github saying "muy fav place iz gone!".

Maybe show a toast or something. There are other occurences where this would be helpful as well, for example when someone tries to launch an app or shortcut that is no longer available.

Sir-Photch:
> Also, we might want to refresh some items again at a later point. After all, searchables can live pretty long (several days), if they are in favorites.

> And there's the question of at what level we implement the caching. Do we just leave it to each implementation and let it decide by its own whether and for how long to cache updated data?

Right now every time the details of a `{Grid, List}Item` are shown, values are refreshed if the closure of `UpdatableSearchable` is not null, and if its `timestamp` is older than one hour. So the whether-to-refresh-or-not logic is kept only in `SearchableItemVM`. 

Also, I've changed it such that every location created by `updatedSelf` will also have this closure set; In other words: "descendants" of deserialized locations will always be able to update themselves.

> In that case I would make it a regular member function of the interface, not a property.

I'm pretty sure that we need to capture `osmRepository` of `OsmLocationDeserializer` in a closure since otherwise we would have to inject it into `data class OsmLocation`.

data/openstreetmaps/src/main/java/de/mm20/launcher2/openstreetmaps/OsmSerialization.kt
```diff
@@ -0,0 +1,116 @@
...
+        return OsmLocation(
+            id = id,
+            latitude = json.getDouble("lat"),
+            longitude = json.getDouble("lon"),
+            category = json.getString("category").runCatching { LocationCategory.valueOf(this) }
+                .getOrNull(),
+            label = json.getString("label"),
+            street = json.optString("street"),
+            houseNumber = json.optString("houseNumber"),
+            openingSchedule = null,
```
Sir-Photch, line 77: Any reason you left this as null? The code is right below and lgtm.

MM2-0: Oops, I probably forgot about it.

Sir-Photch:
Some screenshots of osm search results with varying settings:

<details>
<summary>General result view</summary>



</details>

<details>
<summary>Map view, light mode, unthemed</summary>

- User location is marked green (altonaer straße) and POI is marked with orange outline, if enabled in settings
- both indicators are pulsating





</details>

<details>
<summary>Map view, dark mode, unthemed</summary>



</details>

<details>
<summary>Map view, light mode, themed</summary>



</details>

<details>
<summary>Map view, dark mode, themed</summary>



</details>

Any comments on the UI part? Right now, I am not really satisfied with the "theming" (when activated), since it really messes up the color palette of the map instead of actually making it look good w.r.t. the current material you theme. But there is only so much you can do with color matrices, and in the end, it is optional and looks fine with some color combinations IMO.

I keep the option to show user location in order to set the POI into context; Otherwise, it just shows the POI close up with no information on where in a region it is. The map view dynamically redraws its tiles if a closer zoom can be applied.

Issue #603, opened 2023-10-31
Add "Edit Widgets" option to hamburger menu of search bar when button is deactivated


Fxsch:
I think it would make sense to add an option to edit the widgets to the hamburger menu, that is on the right side of the search bar when the `Edit Widgets` button is deactivated.

stardomains3:
I second this. I like not seeing that button on my home screen. But it seems you need to re-add it to edit widgets and delete it after finishing editing the widgets.

PR #649, opened 2023-12-28
Add menu entry "Edit widgets" if button on the widgets list is not shown

strongville:
This small PR adds to the overflow menu on the search bar an entry for "Edit widgets".

This entry is only added if the button on the widget list is hidden.

Closes enhancement #603.


MM2-0:
Can you add a check so that the menu entry is only visible when the search is closed?

strongville:
Done. Let me know if something else is needed.

PR #652, opened 2023-12-31
Use Material You Compat on Android 12+ if flag is enabled

### LINKED ISSUE #524: color pallet bug when closing launcher settingd
Whenever we open launcher settings and change something and then close settings from recent view

The whole launcher reverts to previous color pallet .

But not the one generated by the launcher

And then I have to change the wallpaper to make it go dynamic again

Screenshot_2023-08-31-14-37-27-233_de mm20 launcher2 release

See wallpaper color and launcher color

strongville:
Some customization layers such as One UI or MIUI don't retrieve a meaningful color on wallpaper change (maybe this is just done on manual wallpaper selection, which kinda misses the purpose of automatic theming...), and on previous versions, for example, Android 11, still on One UI, this launcher shows more coherent colors most of the time.

This PR suggests a new settings flag under "Appearance" to ignore system colors on API 31+ and use same color quantization scheme for Android 8+, technically skipping the conditional that asks system its colors.

Could close #524 or related if it is merged.


strongville:
Wit the refactor made for the settings data, integrate this flag onto migration classes no longer makes sense.
This PR was updated to reflect that.

PR #673, opened 2024-01-24
Added flags "Show seconds" and "Use theme color" for clock widgets

strongville:
This PR does several things at once, all related about the clock widget and its styles:

- Adds two new setting flags for all clock styles:
  - "Show seconds", applying only on default size (non-compact), found only on Orbit style, and
  - "Use theme color", using Material theme colors as accent, found only on Bold style Material You variant
- Adds all adjustments needed for all existing styles to use these two new flags.
- Removes MDY variant for Bold style: applying Material You variant and "Use theme color" settings produces same result on Default variant.
- Adds a new time provider only for the clock, that behaves the same, updating every minute if "Show seconds" is disabled for the clock, but updates using a Handler running as close as possible at millisecond zero if needed.
- New clock widget style "Display", inspired on clocks with seven-segment LED displays.


Sir-Photch:
Yes! The `:` even blink 👍🏼

strongville:
With the new commits, the issue of having icons and other stuff being queried every second, is now solved creating a new time provider (`ProvideClockTime`) that is only used on the `ClockWidget` and `WatchFaceSelector` classes; this provider can update every second or every minute according to the `Show seconds` setting.

The previous provider (`ProvideCurrentTime`) was reverted to the previous state (updating values according only to the system broadcast events `ACTION_TIME_TICK` and `ACTION_TIME_CHANGED`, every minute at most) and it's now composed a little bit more deeply into the tree (next to the Scaffolds, the only place in which is currently queried).

app/ui/src/main/java/de/mm20/launcher2/ui/component/ShapedLauncherIcon.kt
```diff
@@ -250,7 +250,7 @@ fun ShapedLauncherIcon(
                     _badge.progress?.let {
                         val progress by animateFloatAsState(it)
                         CircularProgressIndicator(
-                            modifier = Modifier.fillMaxSize(),
+                            modifier = Modifier.fillMaxSize(0.8f),
```
MM2-0, line 253: why?

strongville: Sorry, not intended for this patchset, it was to make clearer a progress circle on notificacion dot separating the progress from the border.

PR #707, opened 2024-02-22
Add screnshots in readme.md

Drsheppard01:
#706


TimoNotThy:
Maybe make the calender event in english. You could make it something along the lines of `update Kvaesitso`.
I would also suggest showing mainly FOSS apps.
Try to show multiple layouts, not only the one that fills the screen with the clock to show what's possible.

PR #709, opened 2024-02-23
Implement search query validation in WebsiteRepository

Sir-Photch:
This PR addresses an issue mentioned some time ago on Telegram where previously, any search query (that then gets prepended with `https://`) will be sent over the network, since `URLUtil.isValidURL(url)` returns `true` for `https://foobar`. This does, strictly speaking, leak search queries to DNS when website search is enabled.

The proposed changes check for a valid domain (with top-level domain) to be present in the search query, with IDN support for the domain part.

### DIFF
--- a/data/websites/src/main/java/de/mm20/launcher2/websites/WebsiteRepository.kt
+++ b/data/websites/src/main/java/de/mm20/launcher2/websites/WebsiteRepository.kt
@@ -19,11 +19,19 @@ import okhttp3.Request
 import org.jsoup.Jsoup
 import org.jsoup.UncheckedIOException
 import java.io.IOException
+import java.net.IDN
 import java.net.MalformedURLException
 import java.net.URISyntaxException
 import java.net.URL
 import java.util.concurrent.TimeUnit
 
+// https://code.whatever.social/questions/10306690/what-is-a-regular-expression-which-will-match-a-valid-domain-name-without-a-subd#26987741
+private val DomainRegex by lazy {
+    Regex(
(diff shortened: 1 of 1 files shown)


MM2-0:
> since `URLUtil.isValidURL(url)` returns true for `https://foobar`

Which is correct because `https://foobar` is a perfectly valid URL. And it is not even that uncommon. I, for example, have a Node-RED instance at `http://raspberrypi:1880`.

I agree that it's bad to leak queries to DNS providers, but at least as far as I understand your code, it doesn't recognize perfectly valid URLs such as:

- http://raspberrypi
- https://my.server.tld:8080
- http://192.168.178.1
- http://[2001:0db8:85a3:08d3::0370:7344]/
- https://subdomain.domain.co.uk
- http://user:password@foobar:8000

Which is the reason that I went for a "try everything, we can't know if it's a valid URL" solution. I think it would be best to keep that approach as long as the query starts with a valid protocol, so that there is a way to opt-in to DNS lookup. Apart from that, your PR look fine.

Sir-Photch:
I see your motivation and I think that treating prefixed `http://` or `https://` as "I know what I am doing" options to directly request the query is a good idea.

Anyway, I doubt that most people seriously use the launcher search for IP addresses, specify ports or authentication.

> I, for example, have a Node-RED instance at http://raspberrypi:1880.

I assume this is a DNS rewrite in your LAN? Not regarding the port, allowing domains that do not have some kind of top-level domain (and every internet website has to come with a tld, right?) will return to the initial problem.

That's why I suggest to keep the domain validation in case there is no protocol specified:
https://regex101.com/r/ZGkwiQ/1

Sir-Photch:
All of these edge-cases work by prefixing the protocol. Let me know if you want this code to do something else.

PR #745, opened 2024-03-27
translations fixes

zxkmm:
the "fixed" should translate into "static/const" instead of "corrected/revised" 
it's a polysemous issue

---
there seems no dev or next branch, so i pointed to main. lemme know if i should point to another branch.
Thx!

### DIFF
--- a/core/i18n/src/main/res/values-zh-rCN/strings.xml
+++ b/core/i18n/src/main/res/values-zh-rCN/strings.xml
@@ -555,9 +555,9 @@
     <string name="gesture_failed_message">您已执行“%1$s”手势. 此手势当前设置为触发“%2$s”操作. 然而, 由于以下原因, 无法执行该操作:</string>
     <string name="missing_permission_accessibility_gesture_failed">需要启用启动器的无障碍服务才能执行此操作.</string>
     <string name="missing_permission_accessibility_gesture_settings">这个动作要求开启无障碍服务.</string>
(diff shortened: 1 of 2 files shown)


MM2-0:
Thanks, merged it. But FYI the best place to submit improvements to the translation would be our translation platform (i18n.mm20.de).

zxkmm:
I see, thanks!
BTW I'm planning to implement the "first Latin char of pinyin search" feat that used to talked but not implemented in #8
I'm planning to directly join the "first lettered" string on the back or front of the return string of romanize func. But I need to know if this implementation is acceptable to you (it looks not elegant).
Cuz it sounds not elegant, but it's high performance I think.

MM2-0:
> I see, thanks! BTW I'm planning to implement the "first Latin char of pinyin search" feat that used to talked but not implemented in #8 I'm planning to directly join the "first lettered" string on the back or front of the return string of romanize func. But I need to know if this implementation is acceptable to you (it looks not elegant). Cuz it sounds not elegant, but it's high performance I think. So i figure i should ask if this is acceptable before i do it. Thx!

This will probably not work without further tweaking of the matching function. Right now, labels and search query are romanized and then fuzzy matched against each other. Your proposed solution would impact that (the invisible suffix would lead to a lower score, and there might also be false positives).
I think it would be better to make it a separate comparison, and only allow literal (substring) matches for that comparison. And maybe it would be best to  disable fuzzy matching altogether for queries that are shorter than 3 letters.

Issue #760, opened 2024-04-09
[Feature proposal] Integrate public transport departure times for respective `Location` search results


Sir-Photch:
Now that the OSM search provider is merged, my next idea is to integrate departure times for public transport search results.

To be specific, I picture the use case to be: I search for a bus stop in my area / have one marked as my favorite and can see live departure times for the next 3 (arbitrary number) connections stopping at that station, in the same `LocationItem` view.

Especially for local transport (not inter-regional, but subways, streetcars and busses in cities)
I picture this to be pretty handy, since looking up when the next bus for your daily commute comes might allow you (or me, because I am the proposer, hehe) to stay in bed just a little longer :)

There just happens to exist a java library that wraps all the different available APIs of public transport providers: [schildbach/public-transport-enabler](https://github.com/schildbach/public-transport-enabler). The issue there is that this would require juggling of more API keys, if the APIs that are to be supported require one. (Some need keys, others don't. I have yet to find out which.)

Regarding how this could be integrated into the codebase, I see two (three) possibilities from the top of my head:
1. Add another module, just as OSM, that implements `SearchableRepository<PublicTransportStop>`. This would probably integrate easiest in a sense of minimal code change, but causes search strings to be passed to (possibly) closed APIs (privacy issue), and would need matching of `Location`s and `PublicTransportStop`s afterwards, if both are to be displayed in the same view.
2. A different module. But in `SearchService`, only pass search results as queries to this module that are bus-, subway-, streetcar stops and were returned by a `SearchableRepository<Location>`. This would allow to reuse a PublicTransport module between different implementations of `SearchableRepository<Location>`. Downside here is that stops would have to be found by the `SearchableRepository<Location>` first. Where I am from, OSM is pretty complete in cities, but this might not apply to other parts of the world. Otherwise, we'd need a seperate `PublicTransportStopItem` view, which seems unnecessary to me.

Variant two would work nicely if there will ever be something like a google-maps search plugin. (Or any other search provider, for that matter.) But AFAIK, google maps have their own public transport information? Not sure though if they expose that via their APIs; After all, I'm more the OSM type of guy anyway. If other location search providers come with their own information, we could also:

3. Add PublicTransportEnabler as a submodule of the OSM repository, so code further up the chain only sees the `Location` interface and may check something like `if (loc is PublicTransportStop) ...` given that `PublicTransportStop` may be another interface that implements `Location`.

[...]

PR #769, opened 2024-04-20
Add initial support for public transport plugins

Sir-Photch:
As suggested in #760.

Submitting this as draft for you to comment on the way I integrated all the plugin matters.
I've decided to "upgrade" existing `Location`-Results by passing them to the new `PublicTransportStopRepository`, which then returns a wrapper class `PluginPublicTransportStop` of the initial `Location` with some additional data.

Currently missing:
- [ ] Some UI that looks good (lol). Maybe display the departure details next to a smaller map?
- [ ] Settings
- [ ] An actual plugin to test this with. But since the plugin would depend on a stable plugin SDK, I postponed that to some later point in time :)
- [ ] (Serialization) (I judge this not to make sense here, since I see no point to store departure times that possibly expire in a couple of minutes. This will probably be necessary for favourites though)


MM2-0:
> I've decided to "upgrade" existing Location-Results by passing them to the new PublicTransportStopRepository, which then returns a wrapper class PluginPublicTransportStop of the initial Location with some additional data.

I'm not quite sure I understand this approach: doesn't this heavily rely on OSM data being correct? This only works if the OSM stop and whatever provider comes next both agree on the same name and coordinates for the same stop, right? And it also relies on the user having turned on location search in the first place.

But I like the idea of extending the existing `Location` interface, and this gave me a new idea: what if we scrap the `PublicTransportStop` interface altogether and simply extend the existing `Location` interface by a `depatureTimes` property? And instead of building a plugin interface for public transport only, make it an interface for places in general?
Of course, this would still leave the question open about what to do with duplicates (we should probably add a way to merge them somehow).

>  (Serialization) (I judge this not to make sense here, since I see no point to store departure times that possibly expire in a couple of minutes. This will probably be necessary for favourites though)

It probably doesn't make sense to store departure times in the database, but it absolutely makes sense to store the general information about the location. I totally see a use case where I pin all the tram stops in my area to favorites and then only need to long press it to see the latest departure times.

Sir-Photch:
> This only works if the OSM stop and whatever provider comes next both agree on the same name and coordinates for the same stop, right?

Currently yes, but as I said, this draft was initially only about the plugin-specifics.

But still, if we already have OSM data, why not use it? Otherwise, we would have to send either:

1. user queries that don't have any intent in looking up public transport stops (for providers that support free-form string queries), or
2. the user's current location directly (for providers that don't)

to closed APIs. Point 1 is the same motivation of #709. Point 2 is, in many cases, the motivation for people to ditch apps, or even stock android completely.

The OSM overpass API is not closed in that sense, right now, we can change the API endpoint in the settings, and [everyone is free to set up their own instance](https://overpass-api.de/no_frills.html). And then, if we already have OSM data, we only need to send locations that have been classified as a public transport stop by OSM, so we can protect the user's privacy. If there are any stops that are not classified correctly by OSM, then that is an OSM bug. And in that case, users can find motivation to correct OSM, which would enable Kvaesitso to give back to one of its data providers. (Open Source / Open Data argument, really.)

Right now, my idea is to match on both `latitude`/`longitude` and name, or either of them, based on some heuristic. I agree that there is room to improve; I will do so when the plugin API is agreed on. Actually, my idea would have been to pass the whole `Location`-instance to the plugins, for them to decide what to do with the information. But just while coming up with this I realized that inter-app communication apparently only uses strings and integers. 😅 

> And it also relies on the user having turned on location search in the first place.

Yes, because public transport stops are locations.

> It probably doesn't make sense to store departure times in the database, but it absolutely makes sense to store the general information about the location.

Hm, if we wrap an instance of `OsmLocation`, isn't that already getting serialized? We'd only need to query the public transport repository when the user opens up a pinned location that qualifies as a transport stop.

MM2-0:
We shouldn't rely on OSM being the only data provider for places data. At some point, there will be a plugin API for every content type, and then someone could write a Google Maps / Tripadvisor / Yelp / Foursquare plugin. Now, there is little sense in having more than one places provider enabled, so users in this scenario would only enable one of them at a time.
Maybe someone favors Tripadvisor for its more accurate user ratings (side note: we should add a field for user ratings of a place), but unfortunately it lacks public transport stops entirely.
Or maybe, someone isn't interested in places at all and only ever wants to search for public transport locations. In this case we would unecessarily leak the location to OSM servers (and load their servers with an unnecessary request).

> Otherwise, we would have to send either:  queries that don't have any intent in looking up public transport stops

We have to leak some data anyways. I'd say, send it to the plugin, let the plugin decide what to do with it. For example, a plugin for public transport in London could dismiss any request for user locations outside of London.

Also, I'm currently working on a filter system (it is already partially available in Nightly builds), which allows you to toggle search categories on demand and makes search requests offline by default. As long as you don't enable the "online results" filter, no request will ever be sent to a server. And if you aren't interested in places results, you can simply turn them off temporarily. I will also add an option to configure the default filter so that if you rarely are interested in places, you could turn them of by default and temporarily enable them if you need them.

Sir-Photch:
I'll open another PR for location plugins

PR #772, opened 2024-04-21
Add support for location plugins

Sir-Photch:
Successor to #769.


Sir-Photch:
This should be reviewable now.

I have added some UI for the departures with mock data in debug mode:

<details>
<summary>UI demo</summary>

### With map view:

The `LazyColumn` has a bounce animation if it is scrollable, to indicate to the user that there are more entries further down.

[Screen_recording_20240505_135903.webm](https://github.com/MM2-0/Kvaesitso/assets/47949835/26f4c302-97b3-4995-a9fd-99a4381fc75d)

### Without map view:

This one doesn't look as good, there is lots of unused space (I mainly focused on the view that includes the map)



</details>

I am wondering if the background of LineType-Icon and LineName should vary between line-types: blue for busses, red for trains, etc. This might not fit very well with the material theming though. (Maybe blend the colors?)

Right now, scrolling the departures works only if the `LocationItem` is opened up from the favourites, in case it is shown as a search result, the scrolling is somehow inhibited. (`Modifier.consumeAllScrolling()` also doesn't do anything here.) Do you have an idea why this could be?

MM2-0:
> This one doesn't look as good, there is lots of unused space (I mainly focused on the view that includes the map)

Swap the destination and departure columns , make destination fill the available space

> I am wondering if the background of LineType-Icon and LineName should vary between line-types: blue for busses, red for trains, etc. This might not fit very well with the material theming though. (Maybe blend the colors?)

It would make sense to apply the line colors that the transport provider uses, usually lines are color-coded in some way. And to make them fit in, you can adjust the tone of the color (light mode: 40 for container, 100 for content, dark mode: 80 for container, 20 for content). If that still looks bad, you can try to blend them, like how the [WeatherIconColors](https://github.com/MM2-0/Kvaesitso/blob/900e39c6805073aa072ca12c0513308707f5193d/app/ui/src/main/java/de/mm20/launcher2/ui/component/weather/WeatherIconColors.kt) are slightly adjusted to match the theme.

MM2-0:
I'm currently trying to find a good design for the places results, but it's just so hard to fit all the information into the constrained space we have.

Which one do you like better, left or right?
Screenshot_20240505_145500

(for other places, opening hours would be listed in place of the departure times)

Sir-Photch:
Both look nice, I have two ideas:

- Big variant for `LocationItem`s that were opened from favorites, small one for search results (or vice-versa)
- Small variant by default, tapping on map animates to larger variant (since you now have a button for navigation)

By the way, how do you come up with these mock-ups? Figma?

MM2-0:
> Small variant by default, tapping on map animates to larger variant (since you now have a button for navigation)

I was considering this as well. I think it comes down to the question whether the small map is actually usable. Because if it's not, there is no point in showing it at all.


> Also, where did the direction arrow go? To the user location indicator on the map?

Yes, the user indication now shows the direction you're looking at. I think these two information belong together so it makes sense to have them in one place.

> I spent quite some time figuring that out :')

It will still be visible in the compact (list) view. And I will bring it back for the mapless variant.

> What do you think about using icons for the line type, like I have it currently implemented?

Yeah, makes sense. My mockup has no indication for the type of transport. But I would reduce the spacing by a bit.

> And by the way, how do you come up with these mock-ups? Figma?

I use Penpot, which is a Figma alternative that's FOSS and self-hostable.

Sir-Photch:
> I think it comes down to the question whether the small map is actually usable. Because if it's not, there is no point in showing it at all.

You mean not showing a map at all, in any type of location result? I disagree, a map helps to put the location into context. Let's say I am searching for some location and I get two results with the same name. Only a map really helps to distinguish between them. And even the small map is useful to see "ah, that one is downtown" or "further out". This especially applies to the area where you are from, since I argue after living somewhere for a while you know the map of the city to a certain degree. 

> It will still be visible in the compact (list) view. And I will bring it back for the mapless variant.

Hooray! :)

MM2-0:
> And even the small map is useful to see "ah, that one is downtown" or "further out". This especially applies to the area where one is from, since I argue after living somewhere for a while you know the map of the city to a certain degree, where a zoomed-out view with both your location and the POI visible tells you a lot.

But how small can we make the map until that information is no longer conveyed.

By the way, how did you come up with the list for `LocationCategory`? Is it an exhaustive list? I'm wondering whether we should change it to string (to allow plugins to put arbitrary texts there) or to expose the enum to the plugin sdk (and not allow any custom values)

MM2-0:
> No, not at all. I picked the "most common" from https://taginfo.openstreetmap.org/tags. OSM tags consist of key-value pairs like amentity=fast_food, leasure=sports_hall or key=value1;value2:

Do you know if there are some ready-to-use human-readable labels for these tags available somewhere? Maybe I could save my translators some time if I could import these labels from OSM instead of adding and translating them manually.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -477,7 +477,7 @@ fun LocationItem(
                                         Intent.ACTION_VIEW,
                                         Uri.parse("google.navigation:q=${location.latitude},${location.longitude}")
                                     ),
-                                )
+                                ).or(location.launch(context, null))
```
MM2-0, line 659: Afaik `google.navigation:q=lat,lon` is the de facto standard for navigation intents and is supported by several navigation apps (i.e. OsmAnd, HERE We Go). Since there already are two "view on map" buttons (the map itself and the toolbar button), this seems pretty redundant to me and it would be better to remove the chip altogether when there is no navigation app installed.

Sir-Photch: All I can say is that it does not work for me as an OrganicMaps user. Question then is how to determine if there is no app that can handle that intent?

MM2-0: Alright, let's keep it then.

> Question then is how to determine if there is no app that can handle that intent?

You can use [`PackageManager.resolveActivity`](https://developer.android.com/reference/android/content/pm/PackageManager#resolveActivity(android.content.Intent,%20int)) to find an activity that can handle a given intent. It returns `null` if there's none.

Sir-Photch: I've removed the chip now when there is no activity to resolve navigation intents.

core/shared/src/main/java/de/mm20/launcher2/search/location/Departure.kt
```diff
@@ -0,0 +1,26 @@
...
+import de.mm20.launcher2.serialization.DurationSerializer
+import de.mm20.launcher2.serialization.LocalTimeSerializer
+import kotlinx.serialization.Serializable
+import java.time.Duration
+import java.time.LocalTime
+
+@Serializable
+data class Departure(
+    @Serializable(with = LocalTimeSerializer::class)
+    val time: LocalTime,
```
MM2-0, line 14: I think this should be a `LocalDateTime`? Because it refers to one singular trip (including delay), not a recurring trip. Also we should clarify: is this the scheduled time or the real time including delays?

MM2-0:
I have replaced `street` / `houseNumber` with a more generic address object, because there are numerous different address formats around the world, and most APIs probably return the address as one formatted string. But apparently, the OSM API only returns street and house number as individual fields and now we need to find a reliable way to format these data back into an address. So far, I have found [this](https://github.com/placemarkt/address-formatter-java).

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -372,24 +374,10 @@ fun LocationItem(
...
-                            if (isAfternoon) {
-                                afternoonDepartures + morningDepartures
-                            } else {
-                                morningDepartures + afternoonDepartures
-                            }
-                        }
+                    val departures = remember(location.departures) {
+                        location.departures
+                            ?.sortedBy { it.time }
+                    }
```
Sir-Photch, line 377: Does switching to `ZonedDateTime` account for departures past midnight? E.g.: it is 11:45 PM and some departures returned by a provider are past midnight

Sir-Photch:
> But apparently, the OSM API only returns street and house number as individual fields and now we need to find a reliable way to format these data back into an address. So far, I have found [this](https://github.com/placemarkt/address-formatter-java).

Hrm, we could take the first N lines returned by this library for `address{,2,3}` and fill the rest via the [OSM `addr:*` tags](https://wiki.openstreetmap.org/wiki/Key:addr:*), that is:

`Address.city` -> `addr:city`
`Address.state` -> `addr:province` or `addr:state`
`Address.postalCode` -> `addr:postcode`
`Address.country` -> `addr:country`

core/shared/src/main/java/de/mm20/launcher2/search/location/OpeningSchedule.kt
```diff
@@ -0,0 +1,26 @@
...
+    val duration: Duration
+) {
+    override fun toString(): String = "$dayOfWeek $startTime-${startTime.plus(duration)}"
+}
+
+@Serializable
+data class OpeningSchedule(
+    val isTwentyFourSeven: Boolean,
+    val openingHours: ImmutableList<OpeningHours>
+)
```
Sir-Photch: maybe 
```kotlin
sealed class OpeningSchedule {
    data object TwentyFourSeven: OpeningSchedule()
    data class Hours(val openingHours: ImmutableList<OpeningHours>): OpeningSchedule()
}
```
as they should be mutually exclusive?

MM2-0: Yes, good point (but I would use a `sealed interface`). Though that would probably break backwards compatibility with already stored locations, but I think that's okay since the item is updated anyways as soon as the user views it.
Also I have replaced `ImmutableList` with a standard `List` because it doesn't support serialization.

MM2-0:
I think we should split the category field into an (enum) icon field and a (string) category field. I'm currently writing a Foursquare plugin and I feel like a lot of information is lost by coercing the categories into the few available enum values.

For reference, these are the place categories that are available in Foursquare: https://docs.foursquare.com/data-products/docs/categories

And for the icons enum, I think the Google Maps places API has a pretty good list of location types: https://developers.google.com/maps/documentation/places/web-service/supported_types

data/locations/src/main/java/de/mm20/launcher2/locations/LocationSerialization.kt
```diff
@@ -0,0 +1,228 @@
...
+    return if (json.optBoolean("isTwentyFourSeven")) {
+        OpeningSchedule.TwentyFourSeven
+    } else {
+        OpeningSchedule.Hours(
+            openingHours = json.optJSONArray("openingHours")?.let {
+                getOpeningHours(it)
+            } ?: persistentListOf()
+        )
+    }
+}
```
Sir-Photch, line 86: Dead code, serialization happens in `:core:shared`, right?

core/shared/src/main/java/de/mm20/launcher2/serialization/OpeningScheduleSerializer.kt
```diff
@@ -0,0 +1,52 @@
...
+                    jsonElement.jsonObject["openingHours"]!!
+                )
+            )
+        }
+
+        // fallback in case we receive data which was serialized before introducing OpeningScheduleSerializer.
+        // here openingHours is empty, in which case it has to be 24/7, else it would indicate corrupted data.
+        return OpeningSchedule.TwentyFourSeven
+    }
+}
```
MM2-0, line 52: kotlinx.serialization should support polymorphism by default, no?
See https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md

data/locations/src/main/java/de/mm20/launcher2/locations/LocationSerialization.kt
MM2-0: Yes, it's all handled by `kotlinx.serialization`

Sir-Photch:
> I think we should split the category field into an (enum) icon field and a (string) category field. I'm currently writing a Foursquare plugin and I feel like a lot of information is lost by coercing the categories into the few available enum values.

How do you plan on localizing these free-form strings in that case? I am referring to the mappings you have added lately for the location type to be shown as text in the search results.

Also, there would need to be some mapping anyway between some known strings and the icon field, if it is supposed to be an enum, right?

core/shared/src/main/java/de/mm20/launcher2/serialization/OpeningScheduleSerializer.kt
Sir-Photch: Uhh, I will have a look. Don't know this API 😃

Sir-Photch: Okay so probably yes, it says it right here: https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#sealed-classes

but in that case we don't have backwards-compatability?

MM2-0:
> How do you plan on localizing these free-form strings in that case? I am referring to the mappings you have added lately for the location type to be shown as text in the search results.

Send a `lang` parameter to the plugin, let the plugin worry about it.

> Also, there would need to be some mapping anyway between some known strings and the icon field, if it is supposed to be an enum, right?

Why?

Sir-Photch:
How do you create the mapping from category to icon then

MM2-0:
> How do you create the mapping from category to icon then

I don't, they are two independent fields.

Sir-Photch:
So you'd also have this specified by the plugin?

And it has to be done for OSM still. It sounds like a sensible change though, in that case we could also consider `cuisine:*` for `amenity:restaurant` to choose the right icon for example.

MM2-0:
> So you'd also have this specified by the plugin?

Yes

> And it has to be done for OSM still. It sounds like a sensible change though, in that case we could also consider cuisine:* for amenity:restaurant to choose a more accurate icon I suppose.

Yes. And for the category labels, I would import the strings from OsmAnd, to save our translators some work.

core/shared/src/main/java/de/mm20/launcher2/serialization/OpeningScheduleSerializer.kt
MM2-0: > but in that case we don't have backwards-compatability?

Yes, but for the sake of maintainability, I think that's acceptable. As I mentioned earlier, the data is refreshed as soon as the user is online and opens a location's detail view.

MM2-0:
WIP Foursquare plugin: https://github.com/Kvaesitso/Plugin-Foursquare

Overall, most things seem to be working fine. Transmission of opening hours seems to be broken. Some category names are in English for some reason. And of course, I couldn't test departure times yet.

Sir-Photch:
> WIP Foursquare plugin: https://github.com/Kvaesitso/Plugin-Foursquare

Cool, thanks. I'll use that set up a transportation plugin in the coming days 👍🏼

MM2-0:
Great! Do you know how to setup the plugin SDK in your local maven repo?

MM2-0:
Yes, but then you need to set up the Github Maven repo, which isn't trivial either. Plus you can't fix bugs that you might find.

Here is what you need to do:

1. Set the plugin SDK version to something that ends with `-SNAPSHOT` (I have already done this). `-SNAPSHOT` versions differ from normal versions in that they can be updated without changing the version number.
2. Run `./gradlew publish` in Kvaesitso's root directory. This will build and publish the artifacts to your local maven repo (`$HOME/.m2/repository`)
3. Add `mavenLocal()` to your plugin project repositories.
4. Add the plugin SDK dependency (use the SNAPSHOT version)

If you need to make changes to the SDK:
1. Run `./gradlew publish` again
2. In your plugin project, in Android Studio, open the Gradle sidebar (on the left side). Right click your root project and select "Refresh Gradle Dependencies"

MM2-0:
HERE plugin, that supports both places and departure times: https://github.com/Kvaesitso/Plugin-HERE
(TODO: parse opening hours, parse place icons, add weather provider)

MM2-0:
Finally found a use for pinned tags :D

Screenshot_20240602-190421

Sir-Photch:
WIP (!) public transport plugin: https://github.com/Sir-Photch/KvaesitsoPlugin-PublicTransport

I'm running into an issue where just the `id` param of `get()` is not enough to get a `Location` that has all of its fields populated. Specifically, I can get the updated departures given the `id` of a location, but nothing else. For a "full" location, I'd either need the `label` or latitude / longitude of the location that is supposed to be updated via `get()`. I see two solutions to this problem:

1. pass more information in `GetParams`, but this would not really make sense for any other plugins that don't need this additional information
2. Add another `StorageStrategy`, similar to `Deferred`, but indicating that the returned `Location` of `get()` may contain incomplete information / only information that is relevant for updating. The non-null fields in that `Location` may then be used for updating the launchers copy. Maybe `StorageStrategy.Updatable`? Not sure about that name... Do you have something in mind?

A different idea thats coming to my mind is changing the return-type of `get()` to something like `GetResult<Location>?` which can be one of: `GetResult.Update`, `GetResult.Replace` or `null`, indicating that either:
1. the present copy in the launcher datastore is to be updated with all non-null fields of the encapsulated location
2. the present copy is to be replaced by the encapsulated location
3. the copy is to be removed.

MM2-0:
Or maybe `Deferred` should always call `getUpdate` and `StoreReference` always calls `get`

Sir-Photch:
> And if we use null as "keep the old value", then how do we tell the launcher to unset a value?

Indicate not to merge, but to replace the launchers copy.

> Maybe that's fine though, the launcher has to check for that anyways and filter out invalid results.

Since the label is what a user searches for, I'd just define a set of "minimum required" fields that need to be populated, and which need to be set for locations returned by `search`, and make `label` one of them.

> Or maybe `Deferred` should always call `getUpdate` and `StoreReference` always calls `get`
> Maybe turn StorageStrategy? into sealed interface with StoreReference and StoreCopy being data objects and `data class Deferred(val mergeResults: Boolean = false)`.

Problem here is that everything returned by `get()` would be treated as merge?

I had a look at how to add
```kt
sealed interface GetResult<TResult> {
    data class Replace<TResult>(val with: TResult) : GetResult<TResult>
    data class Merge<TResult>(val with: TResult) : GetResult<TResult>
}
```
and to change `get()` to
```kt
open suspend fun get(id: String, params: GetParams): GetResult<TResult>? = null
```

where `null` means "delete from launcher store", and the other two variants (of this "enum") indicate on how the launcher is supposed to behave. Problem is, how to represent this as a cursor? Add another column, like, `GetColumns`?

MM2-0:
Implemented the following changes:
- Removed `StorageStrategy.Deferred`. `StoreCopy` now allows the launcher to refresh the stored data at its own discretion ( = when a user long presses an item)
- Added `open suspend fun refresh(item: TResult, params: RefreshParams): TResult?` which is called when storage strategy is `StoreCopy` and the launcher wants to refresh a item

I haven't tested it yet, there may be bugs.

MM2-0:
I think the only thing left to do is fix OSM categories? And then this can be merged?

Sir-Photch:
> I think the only thing left to do is fix OSM categories? And then this can be merged?

I suppose so.

One more thing though, is the `Attribution` (supposed to) be used in the UI, or shown to the user in any way? I could use this to display which public transport network the location originated from.

MM2-0:
> One more thing though, is the `Attribution` (supposed to) be used in the UI, or shown to the user in any way?

Oh, true. Some providers legally require this.

Sir-Photch:
> I think the only thing left to do is fix OSM categories? And then this can be merged?

And, actually, `address` as well. 🙃 
I'll have a look at `category` though.

MM2-0:
Yet another breaking change in the plugin API. Sorry about that, make sure to update launcher and plugin sdk.

Sir-Photch:
No problem... Actually, one problem: getting a localized string for `OsmLocation.category` requires a `AndroidContext`, doesn't it? How do we pass / inject it at down to that level?

MM2-0:
> Actually, one unrelated problem to that: getting a localized string for `OsmLocation.category` requires a `AndroidContext`, doesn't it? How do we pass / inject it down to that level?

`LocationsRepository` has access to context which you can pass to the constructor of `OsmLocationProvider`.

Oh, by the way, if you add strings for categories, please add them to `poi.xml` and (if possible) copy them from https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/res/values/phrases.xml. I will later import OsmAnd into my Weblate instance and use Weblate's auto translation feature to synchronize the translations automatically.

Sir-Photch:
I've added some tag matching to categorize the OSM locations. However, there seems to be an issue where osm locations that were already marked as favorite are not updating their icons to the now categorized one...

MM2-0:
> This seems to alleviate itself when the launcher is restarted (?)

They should update when the item is refreshed. So it should fix itself when the user taps an item (?)

> ...Aaand I just noticed that you've asked me to copy relevant localization strings over from OsmAnd. Sorry, I will clean them up tomorrow!

All good, and thanks for fixing the categories

Sir-Photch:
> They should update when the item is refreshed. So it should fix itself when the user taps an item (?)

I can't say that this was the case. But now I cannnot test that anymore because the locations I had marked as favorite in the emulator got replaced (by me). So this could be an issue once this update lands and people already have locations saved?

MM2-0:
> I can't say that this was the case. But now I cannnot test that anymore because the locations I had marked as favorite in the emulator got replaced (by me). So this could be an issue once this update lands and people already have locations saved?

I just tested it:
- The category is updated as soon as I long press a pinned location
- The icon is updated when the launcher is restarted. The reason for that is, that the launcher caches icons, so the change isn't reflected immediately

Sir-Photch:
> The category is updated as soon as I long press a pinned location

Yes, category works fine. But is long pressing now required for refreshing?

MM2-0:
> Yes, category works fine. But is long pressing now required for refreshing?

Always has been?

> Also, I had a look at OsmAnd strings; They use a different naming scheme.
Instead of poi_category_park it is just poi_park. The strings that we already had use the former format. In this case, I wouldn't change it for consistency sake

Alright. Gotta move them back to `strings.xml` then.

Sir-Photch:
> Always has been?

Oh, I thought that locations always refresh themselves if they are opened up and the data is older than (now) 1 minute

I have added some parsing for `OsmLocation.address`, based on a library thats based on the one you suggested. The latter used java that was incompatible with android, which is probably why this kotlin "port" exists.

That being said, it is a unversioned library with version code set to `-SNAPSHOT`. I'm not sure if you are okay with this? The repo hasn't seen any commits since two years though, so it might be considered "stable": https://github.com/woheller69/AndroidAddressFormatter

MM2-0:
> Oh, I thought that locations always refresh themselves if they are opened up and the data is older than (now) 1 minute

Oh yeah, right. That's what I meant. You are right, single tap leads to the same result.

> The latter used java that was incompatible with android, which is probably why this kotlin "port" exists.

Like what?

> That being said, it is a unversioned library with version code set to -SNAPSHOT. I'm not sure if you are okay with this? The repo hasn't seen any commits since two years though, so it might be considered "stable": https://github.com/woheller69/AndroidAddressFormatter

Better fork it (copy the source code to a new module in `libs`). To avoid surprises, like breaking API or license changes.

Sir-Photch:
> Like what?

https://github.com/placemarkt/address-formatter-java/blob/77bed8a47b197097b41f9edaf304333161028a4e/src/main/java/net/placemarkt/AddressFormatter.java#L451

PR #900, opened 2024-06-30
Migrate osm opening hours parsing to external library

### LINKED ISSUE #860: Chore: migrate OSM parsing of opening-hours to external library
I have come accross some cases where our current parsing of the OSM `opening_hours` tag is not correct. While digging into the issue, I discovered that in the meantime of merging the OSM PR, some new libraries for just that have popped up. One of them is https://github.com/westnordost/osm-opening-hours which is also used in https://github.com/streetcomplete/StreetComplete.

I suggest migrating to that library instead of re-inventing the wheel, since also, our implementation does not cover month- or season-specific opening hours. What do you think?

Also, would you prefer this change to be a seperate PR or can I just include it in #772?

Sir-Photch:
closes #860.
I have also added tests with JUnit. Not sure if you'd still want new test implementations to base on a Java-only testing library, but I figured since e.g. tests of `:core:ktx` also use JUnit I'd use that.

There are some TODOs left because I somehow underestimated how big of a raging dumpster fire the [`opening_hours` spec](https://wiki.openstreetmap.org/wiki/Key:opening_hours) actually is. ("sunrise to sunset" is a thing!)


MM2-0:
> I have also added tests with JUnit. Not sure if you'd still want new test implementations to base on a Java-only testing library, but I figured since e.g. tests of :core:ktx also use JUnit I'd use that.

Afaik Android officially still only supports JUnit 4 as a testing library, so I'm fine with that.

 > There are some TODOs left because I somehow underestimated how big of a raging dumpster fire the [opening_hours spec](https://wiki.openstreetmap.org/wiki/Key:opening_hours) actually is. ("sunrise to sunset" is a thing!)

Oh no! :D

I'd like to publish a release today, so this will probably have to wait for the next version if that's alright?

Sir-Photch:
> I'd like to publish a release today, so this will probably have to wait for the next version if that's alright?

Sure, no problem. I'm not in a hurry :)

Sir-Photch:
Hey there, it's been a while but finally I found some time to have a look at this again.
I made up my mind and now consider support for the whole opening_hours_spec to be out of scope. Still, this PR adds support for more types of opening_hours than those currently available. I've added tests for parsing, feel free to have a look and tell me whether or not you think that is enough coverage... Anyway, I consider this to be "ready for review".

gradle/libs.versions.toml
```diff
@@ -138,6 +141,9 @@ junit = { group = "junit", name = "junit", version.ref = "junit" }
 androidx-junit = { group = "androidx.test.ext", name = "junit", version.ref = "junitVersion" }
 androidx-espresso-core = { group = "androidx.test.espresso", name = "espresso-core", version.ref = "espressoCore" }
 
+osmopeninghours = { group = "de.westnordost", name = "osm-opening-hours", version.ref = "osmOpeningHours" }
+testng = { group = "org.testng", name = "testng", version.ref = "testng" }
```
MM2-0, line 145: Is this library used anywhere?

Sir-Photch: It is not! :b

PR #958, opened 2024-07-19
Query search against application package name in addition to label.

CepGamer:
When searching apps, compare query against the original app name as well as label to help with i18n. For example, Google suite apps, such as Google Maps, have localised labels in different languages. Google Maps name in Russian is `Карты`, which cause it to not be shown when querying for `map`. Current behaviour:

image

Expected behaviour (after this pull request):
image

### DIFF
--- a/data/applications/src/main/java/de/mm20/launcher2/applications/AppRepository.kt
+++ b/data/applications/src/main/java/de/mm20/launcher2/applications/AppRepository.kt
@@ -222,7 +222,7 @@ internal class AppRepositoryImpl(
                     appResults.addAll(apps)
                 } else {
                     appResults.addAll(apps.filter {
-                        matches(it.label, query)
+                        matches(it.label, query) || matches(it.componentName.packageName.split('.').last(), query)
                     })
 
                     val componentName = ComponentName.unflattenFromString(query)


data/applications/src/main/java/de/mm20/launcher2/applications/AppRepository.kt
```diff
@@ -222,7 +222,7 @@ internal class AppRepositoryImpl(
                     appResults.addAll(apps)
                 } else {
                     appResults.addAll(apps.filter {
-                        matches(it.label, query)
+                        matches(it.label, query) || matches(it.componentName.packageName.split('.').last(), query)
```
MM2-0, line 225: why only the last part?

CepGamer: Because of all the `com` prefixes and company names, e.g. thus any `com` search queries will break.

I'm thinking about using English (or default) label instead by loading context, what do you think? Although non-English apps may not be indexed, however in that situation it may not be a problem

MM2-0: Package names in general are pretty inconsistent (this is the reason I don't use them already). Loading the English label is certainly the better approach, as long as you cache the English label (just add an internal field to `LauncherApp` and store it alongside the app)

CepGamer: Sorry for late response, but tinkering around with it it looks like loading `Context` with different locale config doesn't actually load different labels for apps, since it's being handled by `PackageManager.getText` which, as I understand, is manufacturer dependent. I'll see if I can retrieve it from adb/manifest somehow, but it looks like package name may be an option

jjzhangg:
Is it possible to make the shortcuts on original desktop searchable and selectable？

CepGamer:
> Is it possible to make the shortcuts on original desktop searchable and selectable？

AFAIK the new shortcuts should represent 1 to 1 mapping in Kvaesitso launcher. Can you please elaborate and/or provide example?

PR #1071, opened 2024-09-05
Fix weather sdk providing wrong WeatherLocation

leekleak:
Hi, a fair warning, I didn't actually test if this fixes anything at all because I was unable to compile the module (Any help with that would be appreciated), but I don't see where else the problem could be so it's probably fine.

So yeah, while making my own plugin I found that when the location is set manually and the plugin is using WeatherLocation.Id, the location provided by the sdk has the name and id switched requiring a workaround for in the plugin and this _should_ address that.


MM2-0:
Thanks!

Here are the steps for plugin SDK development:
https://github.com/MM2-0/Kvaesitso/pull/772#issuecomment-2135580060

MM2-0:
The fix has been published in plugin SDK version 2.1.1

PR #1075, opened 2024-09-08
Fix typo in crash report template

FestplattenSchnitzel:


MM2-0:
How did I not notice 😭

PR #1117, opened 2024-10-26
`SearchVM`, `SearchService`: improve search UX by reducing flickering

### LINKED ISSUE #617: Use `mutableStateListOf<Searchable>()` instead of `mutableStateOf<List<Searchable>>` in `SearchVM`
Is there a specific reasoning behind not using a state-list in searchvm over a mutable state that contains a list? If I understand correctly, this would make it possible to `.animatePlacement()` of search result items, by changing the list contents, instead of re-assigning the list that is wrapped by a `mutableState`.

https://github.com/MM2-0/Kvaesitso/blob/5a7600a418a25ca44fc042f7a0d60ed7a2a88565/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt#L59-L69

Even without using `.animatePlacement()`, this way we should be able to just directly modify the lists with the results of `searchService`, instead of creating new lists.

Sir-Photch:
This PR closes #617 and improves search UX by not redrawing search results until there are actual results by any `SearchableRepository`. This "fixes" the "flickering" that would otherwise occur when searching:

https://github.com/user-attachments/assets/723dbf1f-120a-4d1a-ad5a-fa6f3243b304

I also went and replaced `MutableStateOf({Int,Float})` to the respective `Mutable{Int,Float}StateOf` that I found on the way, for good measure.

Now, I haven't tested this much, so I'm not sure whether or not this introduces any regressions. But I think the result of an UI that does not flicker when typing is quite promising. Let me know what you think.

And sorry for that messed up commit-history. :sweat_smile:


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -218,15 +232,6 @@ class SearchVM : ViewModel(), KoinComponent {
                     flowOf(emptyList())
                 }
                 val allApps = searchService.getAllApps()
```
Sir-Photch, line 234: maybe it would make sense to set `previousResults` here as well. But which list is correct? There are three, after all. The normal one?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -254,10 +259,11 @@ class SearchVM : ViewModel(), KoinComponent {
...
 
-                        appResults.value = apps
-                        workAppResults.value = workApps
-                        privateSpaceAppResults.value = privateApps
-                        hiddenResults.value = hiddenItems
+                        searchActionResults.clear()
+                        mergeStateLists(appResults, apps)
+                        mergeStateLists(workAppResults, workApps)
+                        mergeStateLists(privateSpaceAppResults, privateApps)
+                        mergeStateLists(hiddenResults, hiddenItems)
```
Sir-Photch, line 266: Maybe it makes sense to set `previousResults` here too?

Issue #1162, opened 2024-12-10
List like app layout


KorigamiK:
We do not have the ability to turn off icons. 
We need 1 column layout with only app names.

Great for minimalist layout and I think would be easy enough to implement:)

zakir786lrk:
It would be great to have setting to change app view as grid or list

KorigamiK:
Pr#1170 already addresses this but hasn't been merged yet

Issue #710, opened 2024-02-23
app drawer list mode


Monirzadeh:
is it possible to add option for app drawer be just a list without icon?
someting like [mlauncher](https://github.com/HeAndroids2Much/mLauncher)

MM2-0:
Added in 1.35

Issue #995, opened 2024-07-26
Niagara-like layout and features 


GGORG0:
Kvaesitso is currently the best and most updated open source non-Launcher3 launcher. It also is very close to Niagara in terms of its workflow. 

It would be cool to implement (optional) features, such as:

- A scrollbar (which shows the letters of the alphabet, like Launcher3) or just all the letters (like Niagara), also with the option to activate it using the left edge, or both (for using with the left and right hand). When scrolling it would highlight all apps starting with that letter, for example by fading all others away or just decreasing their opacity/lightness. Scrolling can also provide haptic feedback 
- A toggle to change the app grid to an app list, with categories for each letter of the alphabet
- Make the scrollbar (from bullet point #1) immediately accessible on the homescreen (maybe it's hidden but still can be activated by tapping the edge of the screen) - as soon as you tap it/start scrolling, the app grid/list would appear.

I'd be happy if even one of these gets created (I think the scrollbar fits with the current Kvaesitso workflow very well too)

I think Niagara has the best launcher workflow, but:
1. It's not open source
2. Has a very limiting free plan
3. Is missing a lot of customization features that Kvaesitso has
4. Its search engine is very lacking

Aj-style:
Why not use niagra then ?

GGORG0:
> Why not use niagra then ?

read.

> I think Niagara has the best launcher workflow, but:
> 
> 1. It's not open source
> 2. Has a very limiting free plan
> 3. Is missing a lot of customization features that Kvaesitso has
> 4. Its search engine is very lacking

acress1:
Hi 
It would still be awesome to have this feature. Is it planned ?

Thank you very much.

lake-effect:
I think this would be substantial work but worth it. Right now the app list as implemented in #1170 only makes it another tile in the scrolled view, but bounding the scrolled view so that you have widgets above and app pages by letter below would work.

PR #1170, opened 2024-12-13
Feat: App list view

KorigamiK:
Addresses #1162 #710

This allows for creating a list like app layout which don't show icons in the app list.

|     |  |
| -------- | ------- |
| Screenshot_20241214_025543  | Screenshot_20241214_025612
    |


KorigamiK:
The list view has some more fixes and looks like this

Screenshot_20241214_130536
 
Screenshot_20241214_133411

KorigamiK:
Oh and #995 too!

KorigamiK:
@MM2-0 would appreciate a review here

mglants:
Could you please also add alphabet to the right side of screen for fast scrolling

KorigamiK:
> Could you please also add alphabet to the right side of screen for fast scrolling 

I don't think it is necessary that the results will be sorted by alphabets when searching

core/preferences/src/main/java/de/mm20/launcher2/preferences/LauncherSettingsData.kt
```diff
@@ -81,6 +80,8 @@ data class LauncherSettingsData internal constructor(
     val gridColumnCount: Int = 5,
     val gridIconSize: Int = 48,
     val gridLabels: Boolean = true,
+    val gridList: Boolean = true,
+    val gridListIcons: Boolean = true,
```
MM2-0, line 84: List shouldn't be the default value

core/preferences/src/main/java/de/mm20/launcher2/preferences/migrations/Migration4.kt
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    override suspend fun migrate(currentData: LauncherSettingsData): LauncherSettingsData {
+        return currentData.copy(
+            schemaVersion = 4,
+            gridList = true,
+            gridListIcons = true,
+        )
+    }
+}
```
MM2-0, line 21: This migration is unnecessary, `gridList` already has a default value which is used when no value exists in the currently saved settings file.

core/preferences/src/main/java/de/mm20/launcher2/preferences/migrations/Migration3.kt
```diff
@@ -15,7 +15,7 @@ class Migration3: DataMigration<LauncherSettingsData> {
         return currentData.copy(
             schemaVersion = 3,
             locationSearchProviders = buildSet {
-                if (currentData.locationSearchEnabled) {
+                if (currentData.locationSearchProviders.isNotEmpty()) {
```
MM2-0, line 18: Why? Because `locationSearchEnabled` is deprecated? But this migration is the only reason that field still exists…
It used to be a simple `true`/`false` flag but after plugins were added, it is now a list of providers. Since we don't know whether the user has disabled location search before, this migration is necessary.

MM2-0:
Implementation looks good so far, but I'm not a fan of how huge these labels are. Is there a reason you chose such a big font?

KorigamiK:
> Implementation looks good so far, but I'm not a fan of how huge these labels are. Is there a reason you chose such a big font?

Thanks. I wanted the font to be like yam launcher which is nice and a little more minimalist. However I do think we can have the font size customisable for others on their preference too. 

sorry for not looking at the requests earlier.

PR #1182, opened 2024-12-19
More unit conversions

TimoNotThy:
I've added some more conversions including the [Planck units](https://en.wikipedia.org/wiki/Planck_units).

### DIFF
--- a/core/i18n/src/main/res/values/units.xml
+++ b/core/i18n/src/main/res/values/units.xml
@@ -43,6 +43,16 @@
         <item quantity="one">millimeter</item>
         <item quantity="other">millimeters</item>
     </plurals>
+    <string name="unit_nanometer_symbol">mm</string>
+    <plurals name="unit_nanometer">
+        <item quantity="one">nanometer</item>
+        <item quantity="other">nanometers</item>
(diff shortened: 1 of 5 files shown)


core/i18n/src/main/res/values/units.xml
```diff
@@ -43,6 +43,16 @@
         <item quantity="one">millimeter</item>
         <item quantity="other">millimeters</item>
     </plurals>
+    <string name="unit_nanometer_symbol">mm</string>
+    <plurals name="unit_nanometer">
+        <item quantity="one">nanometer</item>
+        <item quantity="other">nanometers</item>
+    </plurals>
+    <string name="unit_picometer_symbol">mm</string>
```
MM2-0, line 51: Should be `nm` and `pm`

MM2-0:
I'm not sure about this… can you think of any scenario where anyone would actually need to convert into plank units or are we just polluting the search results list here?

TimoNotThy:
I see what you mean, nm and pm I would keep in the search. Planck units and other misc units in the future would be nice if implemented trough the plug-in system / as a toggle

PR #1231, opened 2025-01-20
Standardize bottom sheets

leekleak:
Greetings, I wanted to fix the bug where the bottom sheet doesn't display under the navigation bar, but got carried away and kinda moved all bottom sheets to the standard ones from compose material3 lib.

This also unintentionally fixed a bunch of other issues:
```

- Fix a small transparent gap between keyboard and sheet
- Fix needing to swipe very fast to expand the sheet
- Fix sheet sometimes blocking user from exiting
- Fixed a lot of button misalignments

Tag edit scaffold:
- Fix edit tag text box not having a label
- Stop user from creating nameless tags


App edit scaffold:
- Icon selection now opens in a separate sheet and that brings transition animations between the menus.

Icon edit scaffold:
- Hide search bar when no icon packs installed
- Hide "Suggestions" tag when there are no suggestions

Weather location scaffold:
- The search bar is now an actual search bar instead of a regular text field

Clock widget edit scaffold:
- Icons on the buttons now match the text size
```

Unfortunately two minor things did not survive this refactor:
- Sheet names. I doubt this is a big issue as the user pretty much always knows why the sheet opened, however it's worth noting.
- Top actions. I moved them to regular buttons where it made sense.

Although I tried to test as thoroughly as I could, I expect there to be something I missed, so all testing efforts will be appreciated.

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/EditTagSheet.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/EditTagSheet.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.offset
@@ -31,8 +32,8 @@ import androidx.compose.material.icons.rounded.Warning
 import androidx.compose.material3.Button
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.OutlinedButton
 import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.Scaffold
 import androidx.compose.material3.SegmentedButton
 import androidx.compose.material3.SegmentedButtonDefaults
 import androidx.compose.material3.SingleChoiceSegmentedButtonRow
@@ -49,6 +50,7 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.drawBehind
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.PathEffect
 import androidx.compose.ui.graphics.drawscope.Stroke
 import androidx.compose.ui.platform.LocalDensity
@@ -77,8 +79,6 @@ fun EditTagSheet(
 ) {
     val viewModel: EditTagSheetVM = viewModel()
 
-    val isCreatingNewTag = tag == null
-
     val density = LocalDensity.current
 
     LaunchedEffect(tag) {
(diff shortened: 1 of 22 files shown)


app/ui/src/main/java/de/mm20/launcher2/ui/settings/searchactions/EditSearchActionSheet.kt
```diff
@@ -114,14 +114,22 @@ fun EditSearchActionSheet(
...
+                        OutlinedButton(
+                            onClick = { onDelete() },
+                            colors = ButtonDefaults.outlinedButtonColors(
+                                contentColor = MaterialTheme.colorScheme.error
+                            )
+                        ) {
+                            Text(stringResource(R.string.menu_delete))
+                        }
+                    }
+
```
MM2-0, line 132: This button is dangerous because it is easy to press it accidentially (it is in a position where you would normally expect a cancel button). I think it isn't needed anymore because the delete action also exists in the overflow menu for each list item.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/EditTagSheet.kt
```diff
@@ -161,55 +126,81 @@ fun CreateNewTagPage(viewModel: EditTagSheetVM, paddingValues: PaddingValues) {
...
             onValueChange = { viewModel.tagName = it }
         )
+
+        Button(
+            modifier = Modifier.align(Alignment.End),
+            enabled = (viewModel.tagName.isNotBlank() && viewModel.page == EditTagSheetPage.CreateTag && !viewModel.tagNameExists)
+                    || (viewModel.page == EditTagSheetPage.PickItems && viewModel.taggedItems.isNotEmpty()),
+            onClick = { viewModel.onClickContinue() }) {
+            Text(stringResource(R.string.action_next))
+        }
```
MM2-0, line 136: It might be beneficial to force expand the sheet here, to make that button visible without scrolling

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/EditTagSheet.kt
```diff
@@ -161,55 +126,81 @@ fun CreateNewTagPage(viewModel: EditTagSheetVM, paddingValues: PaddingValues) {
...
+        modifier = Modifier.padding(paddingValues),
+        containerColor = Color.Transparent,
+        bottomBar = {
+            Box(modifier = Modifier.fillMaxWidth()) {
+                Button(
+                    onClick = { viewModel.closeItemPicker() },
+                    modifier = Modifier.align(Alignment.BottomEnd)) {
+                    Text(stringResource(R.string.action_next))
+                }
+            }
```
MM2-0, line 155: It might be beneficial to force expand the sheet here, to make that button visible without scrolling

Also, I think a bit of semi-transparent background would be beneficial for this button. And block all touch events behind through background.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/notes/NotesWidget.kt
```diff
@@ -463,6 +435,41 @@ fun NoteWidgetConflictResolveSheet(
...
+                    Icon(
+                        Icons.Rounded.CheckCircleOutline,
+                        null,
+                        modifier = Modifier
+                            .padding(end = ButtonDefaults.IconSpacing)
+                            .size(ButtonDefaults.IconSize)
+                    )
+                    Text(stringResource(R.string.note_widget_conflict_keep_selected))
+                }
+            }
```
MM2-0, line 472: <details>

Screenshot_20250120-201801
</details>

leekleak: I'm sorry I'm not sure what's the problem here

MM2-0: The button text shouldn't wrap. But it's not really your fault, I just didn't notice it before. I'm sure the problem existed before on some languages/screen sizes.

MM2-0:
Looks good to me. Is there anything you want to change or can this be merged?

leekleak:
> Looks good to me. Is there anything you want to change or can this be merged?

Codewise nothing, but perhaps unused strings should be removed?

MM2-0:
> Codewise nothing, but perhaps unused strings should be removed?

Good point, but I'd imagine there are a lot of unused resources and it would be easier to remove them all at once, so maybe we do that later, in a new commit.

PR #1232, opened 2025-01-20
Fix search bar transition color

leekleak:
Right now when the search bar is invisible the animation for its background goes like this:
Invisible -> Dark surface -> Light Surface -> Dark surface

That's pretty jarring. With this change it goes:
Invisible -> Dark surface


leekleak:
Will be fixed in #1422

PR #1233, opened 2025-01-20
Widget editor fixes + force resize widgets

leekleak:
Right now when you try to resize a widget to less than 128dp the handles attached to the resizable frame detach which looks weird.

This pr fixes that and also:
- Increases the area from which the left handle can be grabbed from as the system often misinterprets my attempts to grab it as a 'back' gesture
- The widget can be resized by grabbing anywhere on its side as well

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/component/DragResizeHandle.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/component/DragResizeHandle.kt
@@ -73,13 +73,18 @@ fun DragResizeHandle(
         val measuredWidth = this.maxWidth
         val measuredHeight = this.maxHeight
 
-        val actualMaxWidth = if (maxWidth.isUnspecified) measuredWidth else min(measuredWidth, maxWidth)
-        val actualMaxHeight = if (maxHeight.isUnspecified) measuredHeight else min(measuredHeight, maxHeight)
+        val actualMaxWidth = if (maxWidth.isUnspecified || forceResize) measuredWidth
+            else min(measuredWidth, maxWidth)
+        val actualMaxHeight = if (maxHeight.isUnspecified || forceResize) measuredHeight
+            else min(measuredHeight, maxHeight)
         val actualMinWidth = if (forceResize) 40.dp else max(minWidth, 40.dp)
         val actualMinHeight = if (forceResize) 40.dp else max(minHeight, 40.dp)
 
-        if (width !in actualMinWidth..actualMaxWidth || height !in actualMinHeight..actualMaxHeight)
-            onResize(width.coerceIn(actualMinWidth, actualMaxWidth), height.coerceIn(actualMinHeight, actualMaxHeight))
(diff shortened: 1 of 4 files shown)


MM2-0:
I like the idea of making the draggable area fill the entire width but I think the same should apply to the horizontal resizing for consistency.

Also, this breaks the horizontal handle placement for some reason:

<details>

Screenshot_20250122-175820

</details>

leekleak:
> I like the idea of making the draggable area fill the entire width but I think the same should apply to the horizontal resizing for consistency.
> 
> Also, this breaks the horizontal handle placement for some reason (or was that intentional?):
> 
> Screenshot_20250122-175820

Very unintentional 😅. Just poor testing on my behalf. The commit I just pushed should fix that.

Although I also got a bit carried off and added:

- An option to force resize widgets as most of the time widgets can be resized outside of their preset bounds with no problem.
- Vibrations when hitting minimum height and width (only max height and width had them before)
- Hide resize arrows when a widget cannot be resized in a particular direction
- Width indicator at the top left corner now shows the dp when at full width

MM2-0:
> Hide resize arrows when a widget cannot be resized in a particular direction

That works, I guess. I tried to implement that using [`resizeMode`](https://developer.android.com/reference/android/appwidget/AppWidgetProviderInfo#resizeMode) but that turned out to be too restrictive (sometimes, widgets provide some leeway, but not enough to show allow resizing in grid based launchers). Maybe add a comment that we can't use `resizeMode` there

> Width indicator at the top left corner now shows the dp when at full width

That was intentional. 100% actually means full width, it changes its with when you rotate the screen.

leekleak:
> That works, I guess. I tried to implement that using [`resizeMode`](https://developer.android.com/reference/android/appwidget/AppWidgetProviderInfo#resizeMode) but that turned out to be too restrictive (sometimes, widgets provide some leeway, but not enough to show allow resizing in grid based launchers). Maybe add a comment that we can't use `resizeMode` there
I mean now that there's an option to force resize widgets we can use `resizeMode` as well.


> That was intentional. 100% actually means full width, the width changes when you rotate the screen.
My bad -- I didn't know landscape mode was even an option 😖 (Although it looks quite broken on my device)

leekleak:
Probably abandoning this PR as I got really confused about some stuff and I feel like I'd rather not burden myself now. Meanwhile here are some bugs/issues I found as a reference for my future self or someone else:

- If widget is set to 100% width in portrait mode but its max width is shorter than the landscape width of the screen, it will be stretched outside its max width.
- I am not sure if it is a problem per say, maybe there should be a warning for the user if the width of the widget was set in landscape mode and is bigger than the maximum possible width in portrait mode.
- If a widget has minimum width set to zero, it will be set to zero (and thus will be invisible) by default. This happens with some samsung widgets. I feel like the minimum should be around 40dp
- The dimensions label will overflow and overlay over most of the widget when width is small. Should be capped at 2 lines and possibly hidden when there's no room
- The handle also hide a lot of the widget when at small enough sizes
- The other issues mentioned in this thread still stand

PR #1253, opened 2025-01-31
OSM: search by tags "amenity" and "shop" after delocalizing

shtrophic:
As discussed in #1236.


data/locations/src/main/java/de/mm20/launcher2/locations/providers/openstreetmaps/OsmLocationProvider.kt
```diff
@@ -126,12 +129,13 @@ internal class OsmLocationProvider(
...
+        val delocalizedCategory = delocalizeCategoryQuery(query)
+
         val result = awaitAll(
-            // optionally query by "amenity" or "shop" here
-            // if we want to make searching for locations fuzzier
-            // however, this would not account for localized queries like "Bäcker" (shop:bakery)
             Scope.async { searchByTag("name") },
             Scope.async { searchByTag("brand") },
+            Scope.async { delocalizedCategory?.let { searchByTag("amenity", it) } },
+            Scope.async { delocalizedCategory?.let { searchByTag("shop", it) } }
```
MM2-0, line 138: Is there no way to combine these into a single Overpass request?

shtrophic: Yes, there actually is: https://wiki.openstreetmap.org/wiki/Overpass_API/Overpass_QL#Union

I'll refactor shortly...

data/locations/src/main/java/de/mm20/launcher2/locations/providers/openstreetmaps/OsmLocationProvider.kt
```diff
@@ -109,62 +110,68 @@ internal class OsmLocationProvider(
...
-            .toImmutableList()
+        }?.sortedBy {
+            it.distanceTo(userLocation)
+        }?.take(7)?.toImmutableList() ?: emptyList()
     }
+
+    private val poiCategories by lazy {
+        R.string::class.java.declaredFields.mapNotNull {
+            if (it.name.startsWith("poi_category_"))
+                it to it.name.removePrefix("poi_category_")
```
shtrophic, line 162: this is not correct actually. `poi_category_pizza_restaurant` is queried as `amenity=pizza_restaurant`, which is not correct. Ideally, it should be `amenity=restaurant` and `cuisine=pizza`...

shtrophic: Added some (restaurant centric for now, of course :) ) special handling for the names of the resource strings.

So, any resouce string with name `poi_category_asdf_restaurant` will query as `cuisine=asdf` and `amenity=restaurant`. This should be extensible to any other categories that combine tags to make sense, with an requirement for strict resource naming.

PR #1266, opened 2025-02-09
Collapse search objects on query change

leekleak:
Right now whenever you expand something like a contact, it will stay open taking up a lot of space in subsequent searches


MM2-0:
I think this is a better way to fix this: f291c40

PR #1272, opened 2025-02-15
Render icons at resolution matching their size

leekleak:
Currently some icons suffer from blurriness, especially when the app icon size is small. 

That is because those icons are rendered at the app icon size and then scaled to match their actual size. I thought that this was a deliberate choice for performance reasons, but that doesn't really makes sense as when the default app icon size is large, they are rendered large and then downsized, which, obviously, is bad for performance. On the other hand, if the default app icon size is small, the whole process leads to blur. Seems like the worst of both worlds. This changes simply makes it so the icon is rendered at the required resolution to forego any kind of scaling.


aqilc:
@MM2-0 Any thoughts on this?

PR #1281, opened 2025-02-24
Add option to invert favorite tags position

bnason:
Adds an option in the favorites settings to "invert" the position of the tags with respect to the app icons.

The reasoning is that when changing tags that have different number of rows, the tags list jumps around the screen. This makes the tags list position stable.

Current behavior with "Search bar position" set to Bottom:
image

With option checked:
image


shtrophic:
I strongly support this feature ✌🏼

TiemenSch:
This feature/fix is the only thing stopping me from switching to Kvaesitso! It would make the UX so much better and let's the tags double as folders much more neatly. Any chance it can get some priority?

bnason:
@MM2-0 any chance of this being included? I can update the code for the latest version just need to know.

PR #1284, opened 2025-02-28
Fix bottom sheet overlapping status bar

leekleak:
In #1231 I forgot to add top padding to the bottom sheets :/


MM2-0:
I'd almost say that's intentional? It only affects the drag handle, right? It doesn't obstruct any actual content? I'd argue that it looks better when the background surface fills the entire screen, and there isn't a gap at the top of the screen.

leekleak:
Fair point. For me it's only an issue because it looks really weird when the handle overlaps with the camera cut out.

Also, I am not sure if it's a good idea to overlap two interactable elements over each other.

MM2-0:
Yeah, maybe you're right. Pixel Launcher and Google Maps both leave a gap between the bottom sheet and the status bar. The most elegant solution I've seen (I don't remember where) is that the bottom sheet corners slowly morph into rectangular corners and the status bar color changes to match the sheet background color when touch. But I'm not sure if that's worth the effort. So let's just merge this for now.

PR #1285, opened 2025-02-28
Use simple color for icon shape selection

leekleak:
I've found that with certain color schemes it can be quite hard to tell the shapes apart due to poor contrast:

Here's how it looks if instead of the icon there's a simple color:


MM2-0:
Looks much better, thanks!

PR #1286, opened 2025-02-28
Hotfix icon shape selection

leekleak:
Really sorry about that but with #1285 I accidentally removed the ability to actually change your icon shape.


MM2-0:
lol

PR #1294, opened 2025-03-09
LocationItem: localize departure time preview

shtrophic:


core/i18n/src/main/res/values/strings.xml
```diff
@@ -1014,4 +1014,6 @@
     <string name="weather_forecast_humidity">Humidity</string>
     <string name="weather_forecast_wind">Wind</string>
     <string name="weather_forecast_precipitation">Precipitation</string>
+    <string name="departure_time_now">now</string>
+    <string name="departure_time_in_n_minutes">in %1$d minutes</string>
```
MM2-0, line 1018: Should be a plural resource.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/location/LocationItem.kt
```diff
@@ -448,17 +448,20 @@ fun LocationItem(
...
-                                                if (timeLeft < 1) "now" else "in $timeLeft min"
+                                                if (minutesLeft < 1)
+                                                    R.string.departure_time_now to null
+                                                else
+                                                    R.string.departure_time_in_n_minutes to minutesLeft
                                             }
 
                                             Text(
-                                                text = formattedTime,
+                                                text = stringResource(resId).let { if (nMinutes != null) it.format(nMinutes) else it },
```
MM2-0, line 464: I don't understand this construct. Why not keep it as is and replace `"now"` with `context.getString(...)` and `"in $timeLeft min"` with `context.resources.getQuantityString(...)`?

shtrophic: Didn't know about plurals. Will refactor when I find time

core/i18n/src/main/res/values-de/strings.xml
```diff
@@ -813,4 +813,9 @@
     <string name="reset_icon">Symbol zurücksetzen</string>
     <string name="preference_compact_tags">Kompakte Tags</string>
     <string name="preference_compact_tags_summary">Beschriftungen oder Symbole von Tags ausblenden um den von Tags eingenommenen Platz zu reduzieren</string>
+    <string name="departure_time_now">Sofort</string>
```
MM2-0, line 816: "Jetzt"? But I can update it in Weblate

shtrophic: I suppose that's up to taste - the live departure tables at the stops of my place always say "Sofort"

PR #1295, opened 2025-03-09
LocationItem: group departures by line

shtrophic:
Hi,

this PR groups LocationItem departures by line. The motivation behind this is that for a given stop served by many lines, it might be difficult to find the information you're looking for.

Demo:

[Screen_recording_20250309_174430.webm](https://github.com/user-attachments/assets/0ef8456e-d1f1-4a04-8b1c-a099a81a12ae)

UI specifics (especially the question on how to highlight a selected LineFilterChip) are probably up for debate. But otherwise, what do you think?

Also: it probably makes sense to come up with some heuristic that only groups departures by line if the number of `lines * departures[line]` is great enough. Right now, the code falls back to the old UI when there are only departures for one line.


MM2-0:
Makes sense, especially when there are a lot of lines. Two things I have noticed:

- You tap on a button labeled as "U2", but then it shows the departure times for line U3. It would be more intuitive if the line was preselected that is visible in the preview, imho
- The order of lines seems a bit arbitrary? I assume it's sorted by next departure (including past departures), but I think it would be clearer if it was sorted by name and/or line type

> especially the question on how to highlight a selected LineFilterChip

Transparent/outlined chip with colored icon for unselected lines; filled background with monochrome text for the selected line would be my first idea.

shtrophic:
> The order of lines seems a bit arbitrary? I assume it's sorted by next departure (including past departures), but I think it would be clearer if it was sorted by name and/or line type

It is sorted by the number of departures per line, descending.

Otherwise, I have integrated your suggestions:

[Screen_recording_20250314_213747.webm](https://github.com/user-attachments/assets/6ca2e7e8-7273-4db4-9f2e-cdf6ff455f5f)

I've also added horizontal dividers to improve legibility. Now, it looks more like an actual departure table. Let me know what you think.

MM2-0:
> It is sorted by the number of departures per line, descending.

That's confusing as hell. How am I supposed to know the number of departures for the line that I'm looking for?

shtrophic:
I am not sure whether or not we can rely on line names being numeric. Still, something like a simple string-compare after sorting by type should work

MM2-0:
Yeah, I've updated my last comment.

shtrophic:
The only thing that is bugging me is that the width of the filter chips are somehow a clean multiple of the CardOutline width, so an unsuspecting user may not know that the row of filter-chips is scrollable. I'd suggest an animation that initially bounces the list a little when only partially shown?

core/shared/src/main/java/de/mm20/launcher2/search/location/Departure.kt
```diff
@@ -32,16 +32,17 @@ data class Departure(
...
+    Train,
+    RegionalTrain,
+    CommuterTrain,
+    Subway,
     Monorail,
     CableCar,
-    Airplane,
+    Tram,
+    Bus,
 }
```
shtrophic, line 48: this won't break API, amirite? What about ABI? (since the ordinals change)

MM2-0: Afaik Kotlin Serialization uses the name of the enum value, if not specified otherwise, so this shouldn't break anything.

shtrophic:
This is how it looks now:

[Screen_recording_20250327_195032.webm](https://github.com/user-attachments/assets/dd99910a-4b81-4e75-81ce-7dfb12aebbd6)

MM2-0:
> Also: it probably makes sense to come up with some heuristic that only groups departures by line if the number of lines * departures[line] is great enough. Right now, the code falls back to the old UI when there are only departures for one line.

I've decided to always show the grouped UI. Better be consistent.

I've also limited the number of departures to 8 per line and removed the inner scroll column. Nested scrolling feels to finicky.

MM2-0:
I expected this to look terrible, but it's actually alright (but the spacing is off because the chips are scaled down):

PR #1324, opened 2025-03-29
ContactItem improvements

### LINKED ISSUE #163: Make Phone call directly from selecting contact > number
Firstly, One of the best launchers I have used. Thanks a lot, MM2-0!!!!
This brought me proper use for widgets.

Now coming to issue : 
For search, the permission for contacts should serve both Call and SMS.
Please try to make Phone call directly from clicking on number from contact box (search) 
You can make it option to toggle in setting,

The default behavior is, it puts that number in dialer (It takes more than 3 steps, which is not required)
Also add button to message that contact directly, even if we have not started messaging with that contact.

Or else, this feature becomes no use for searching.

For reference, KISS launcher does it. It does by simply showing call or message icon beside contact. That is enough and serves the purpose.

I hope you get my point, let me know if any clarification or if you didn't get me.

Thank you very much MM2-0!

### LINKED ISSUE #1199: Contact names in search results open dialer app instead of calling directly
Contact names in search results open dialer instead of calling directly to that contact. As this launcher is to make everything as easy and fast as possible, this makes it slow to make calls.

### LINKED ISSUE #1304: Contacts search returns two identical phone numbers for all contacts
Contacts search returning number twice

shtrophic:
Closes #1199
Closes #163
Fixes #1304

Also fixes a long-standing bug where any call-activities of other apps (WhatsApp, Signal) wouldn't start because of missing call permissions. Now, the respective activities are only listed in the ContactItem view when we have permission to start them.

This adds the `CALL_PHONE` permission to Kvaesitso. I've "hidden" it behind the ContactsSearch setting. Maybe there is a better place?

Although for #163 it seems like people would like to just tap enter for a contact to be called... However this conflicts `preferDetailsOverLaunch` for `Contact`. Maybe another setting? Then CallOnTap wouldn't be the only Contacts sub-setting :)


app/ui/src/main/java/de/mm20/launcher2/ui/settings/SettingsActivity.kt
```diff
@@ -227,6 +228,9 @@ class SettingsActivity : BaseActivity() {
                                 composable("settings/favorites") {
                                     FavoritesSettingsScreen()
                                 }
+                                composable("settings/contacts") {
```
MM2-0, line 231: Should be `settings/search/contacts` for consistency.

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/contacts/ContactItem.kt
```diff
@@ -295,11 +301,22 @@ fun ContactItem(
                                     app.key
                                 }
                             }
+                            val itemsWithPermission = remember(app) {
+                                app.value.filter {
+                                    // exclude activities we have no permission for
+                                    val resolvedActivityInfo = context.packageManager.queryIntentActivities(
+                                        Intent(Intent.ACTION_VIEW).setDataAndType(it.uri, it.mimeType),
+                                        0
+                                    ).firstOrNull()?.activityInfo ?: return@filter false
```
MM2-0, line 310: `packageManager.queryIntentActivities(…).firstOrNull()` is equivalent to `packageManager.resolveActivity(…)`

zakir786lrk:
There can be separate icons for viewing contact and direct dialing

PR #1342, opened 2025-04-03
Remove phone number formatting

leekleak:
It seems like the recent changes have caused most (all?) local phone numbers to be ignored along with special numbers like 112/911.

Also it seems like the logic for dedublicating phone numbers is wrong for APIs >=31.


shtrophic:
Yeah, sorry for my ignorance. Can you try #1343 maybe?

leekleak:
> Yeah, sorry for my ignorance. Can you try #1343 maybe? That should also handle emergency numbers correctly...
I didn't test much, but it seems to do the same thing.

> The API doesn't say that it needs to be inverted: https://developer.android.com/reference/android/telephony/PhoneNumberUtils#areSamePhoneNumber(java.lang.String,%20java.lang.String,%20java.lang.String)
Yeah, I'm honestly not sure why, but when I was testing this PR, that line had to be inverted to work. On your PR it seems to work fine. Black magic I say.

PR #1355, opened 2025-04-07
Feature: Lock Single Tag for Favorites Widget

rac146:
Wanted to add a feature to this awesome launcher! This isn't quite ready yet but this is my first time diving into the code and would like some feedback/improvements.

This will allow someone to lock a single tag on the favorites widget. Currently if I select a particular tag on the Favorites Widget, it will lose that selection upon app restart or reboot. This new feature will lock it in and can be configured in the widget settings.

If this setting is on, it will also not show the edit button or any other tags in the Favorite Widget. In my opinion this looks really clean visually!

image
image

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/ConfigureWidgetSheet.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/ConfigureWidgetSheet.kt
@@ -34,8 +34,12 @@ import androidx.compose.material.icons.rounded.HelpOutline
 import androidx.compose.material.icons.rounded.Link
 import androidx.compose.material.icons.rounded.LinkOff
 import androidx.compose.material.icons.rounded.OpenInNew
+import androidx.compose.material.icons.rounded.Tag
 import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.CheckboxDefaults
+import androidx.compose.material3.DropdownMenu
+import androidx.compose.material3.DropdownMenuItem
+import androidx.compose.material3.FilterChipDefaults
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
@@ -59,6 +63,7 @@ import androidx.compose.ui.input.pointer.PointerEventPass
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.isUnspecified
@@ -71,7 +76,10 @@ import de.mm20.launcher2.permissions.PermissionGroup
 import de.mm20.launcher2.permissions.PermissionsManager
 import de.mm20.launcher2.plugin.PluginRepository
 import de.mm20.launcher2.plugin.PluginType
+import de.mm20.launcher2.search.Tag
 import de.mm20.launcher2.search.calendar.CalendarListType
+import de.mm20.launcher2.searchable.PinnedLevel
+import de.mm20.launcher2.services.favorites.FavoritesService
 import de.mm20.launcher2.themes.atTone
 import de.mm20.launcher2.ui.R
 import de.mm20.launcher2.ui.base.LocalAppWidgetHost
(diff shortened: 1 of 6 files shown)


app/ui/src/main/java/de/mm20/launcher2/ui/launcher/sheets/ConfigureWidgetSheet.kt
```diff
@@ -184,7 +194,24 @@ fun ColumnScope.ConfigureFavoritesWidget(
...
     onWidgetUpdated: (FavoritesWidget) -> Unit,
 ) {
+    val favoritesService: FavoritesService = get()
+
+    val pinnedTags by remember {
+        favoritesService.getFavorites(
+            includeTypes = listOf("tag"),
+            minPinnedLevel = PinnedLevel.AutomaticallySorted,
+        ) as Flow<List<Tag>>
+    }.collectAsState(emptyList())
```
rac146, line 204: There's probably a better way to do this but currently there's no view model attached so casting to List<Tag> for now

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/favorites/FavoritesWidget.kt
```diff
@@ -56,17 +62,19 @@ fun FavoritesWidget(widget: FavoritesWidget) {
...
+                    selectedTag = selectedTag,
+                    editButton = favoritesEditButton,
+                    reverse = false,
+                    onSelectTag = { viewModel.selectTag(it) },
+                    scrollState = rememberScrollState(),
+                    expanded = tagsExpanded,
+                    compact = compactTags,
+                    onExpand = { viewModel.setTagsExpanded(it) },
+                )
+            }
```
rac146, line 77: dont show any tags/edit if this feature is on

rac146:
Hey @MM2-0 can you check this out and let me know your thoughts?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/favorites/FavoritesWidget.kt
```diff
@@ -41,6 +43,10 @@ fun FavoritesWidget(widget: FavoritesWidget) {
         viewModel.updateWidget(widget)
     }
 
+    if(singleTag) {
+        viewModel.selectTag(singleTagValue);
+    }
+
```
MM2-0, line 49: This can be moved to `FavoritesWidgetVM.updateWidget`, I believe

MM2-0:
Sorry for the late reply. Thanks for the PR, looks good so far, however, I would suggest to add an option to select multiple tags So you can either

- display favorites (with all pinned tags) (the current widget)
- display one tag (which would hide the tag selector bar)
- display two or more tags (which would show the tag selector bar, but without the favorites and the edits buttons

What do you think?

rac146:
Sure that sounds good!

Can you help with the UI design? Trying to come up with the best layout..

Maybe the switch says "Filter Tags" (or something similar) instead of "Set Single Tag"? The dropdown to "Select Tags" can be a multiple select dropdown and then do the logic on one vs multiple tags like you suggest.

Something like:
Edit Button (y/n - only show if Filter tags is N)
Compact Tags (y/n)
Filter Tags (y/n)
Select Tags (Dropdown - only show if Filter Tags is Y)

MM2-0:
> Can you help with the UI design?

Sure, how about something like this?

Start with a segmented button at the top to switch between favorites and tags mode

[ Favorites | Tags] (not sure about the labels)

The options for favorites would be:
- Edit button (y/n)
- Compact tags (y/n)
- The "edit favorites" link

And for tags:
- A drag n drop (?) list of the added tags, with a button to add a new tag
- Compact tags (y/n), only enabled if there are at least two tags

And maybe we can come up with a better name for the widget, because it's no longer strictly a favorites widget

rac146:
Hey @MM2-0 , can you take another look?

This is getting closer - I ended up adding the following features:
- Display Favorites + Pinned tags - default behavior/same as today
- Display Favorites + Custom tags - can change tags and order of tags to be shown next to Favorites
- Display Favorites only - no bottom selection bar, just the favorites
- Display Tags only - show tags only. bottom bar is only shown if more than 1 tag selected

Configuration:
image
- Multi select button for Favorites and Tags (at least 1 must be selected. Defaults to both)
- If no custom tags are selected, it will show all Pinned Tags
- Custom tags can be added and re-ordered with a drag and drop
- I kept most of the other links and settings active to keep it more intuitive. Adding logic around all the buttons started to make it feel busy. The only logic for settings is 'Compact Tags' only visible if 'Tags' is selected

If you can, try this version for a bit and play around with the settings. It feels pretty intuitive overall and easy to configure, but needs a little more polishing. Also not sure how migration would work for current users - might be OK since it defaults to Favorites + Pinned Tags.

I think a good name for this updated widget can be "Shortcuts" (TODO) - what do you think?

rac146:
Hey @MM2-0, did you get a chance to look at this? I've been running this branch for a few months now and has been working great. Just needs some code cleanup and UI approval.

TiemenSch:
Could you include a toggle to put the favorite button and tag selection row within the favorites panel above the favorite results for top-down sorted results and below the results for bottom up sorting? 

That way the favorite/tag select row does not move while switching between tags. Currently the tag selection row becomes a moving target when selecting a tag with a different number of rows of results. 

See https://github.com/MM2-0/Kvaesitso/pull/1281 for an earlier implementation

PR #1411, opened 2025-05-09
fix regressions in parseOpeningHours, ignore holidays

shtrophic:


shtrophic:
Actually, a quick search reveals that there are public APIs for public holidays: https://date.nager.at/Api

Probably overkill for such little benefit though...

MM2-0:
> Actually, a quick search reveals that there are public APIs for public holidays: https://date.nager.at/Api
>
> If you think that integrating this makes sense, I can have a look at that. Maybe overkill though...

Overkill, I'd say

MM2-0:
Can this be merged?

shtrophic:
Yes

Issue #381, opened 2023-05-02
Navigation bar appears even though it is disabled


shtrophic:
It is visible on this video:

https://www.veed.io/view/35b28a3b-8e23-4d34-b962-4de317080720?sharingWidget=true&panel=

Where the navigation-bar still pops up in certain cases, e.g. when anything close to the bottom of the screen is tapped. Also, when unlocking the device, the navigation bar first appears, just to disappear again, which causes the search-bar to "jump".

If you have the swipe-down-gesture set to the notification-center, you can see this when you return back to the home screen.

PR #1422, opened 2025-05-17
Layout rewrite

### LINKED ISSUE #57: Search bar is not completely hidden
Even though I have set the search bar to be hidden, I am still able to touch on the corner and open the settings menu.

search_options

It would be better if I could open this settings menu by long pressing anywhere on the screen, since my search bar is hidden.

### LINKED ISSUE #725: [Feature Request] Allow to choose in which "direction" widgets live.
I think the ability to chose in which way to swipe in order to get to the widgets would be nice. 

For me the widgets would be way more useful if they lived to the side of the home screen. With one swipe, I'd be able to see a whole page of widgets rather than having to slowly scroll down and look for the widget I need.

I hope this makes sense.

### LINKED ISSUE #866: [Feature Request] Ability to remove the whole search bar.
It will help users who want to keep their home screen minimal and who doesn't use the functionality that the search bar provides.

### LINKED ISSUE #993: [Feature Request] Customizable Swipe Up Gesture
Currently, the swipe up gesture opens the widgets, and it's not customizable unlike swipe down/left/right. Please allow binding another gesture to widgets and free up "swiipe up" so that it can be used for other tasks. Opening the app drawer is a common use case for the swipe up gesture.

### LINKED ISSUE #1003: [Feature Request] Allow adjusting swipe threshold distance for gestures
I have to swipe really for for my swipe right and swipe left gestures to perform the action of launching an app.

It would be nice to have a setting to change the distance required to swipe.

(I might make another issue, but it would also be nice to have the option to launch an app with a gesture without the animation showing the app arriving onto the screen from the left or the right. Comparing to KISS launcher which instantly launches the app with no animation when using a swipe gesture.)

### LINKED ISSUE #1164: Configure swipe up gesture to open search instead
Right now there's no way to change the place where widgets are placed. 
I need a way to swipe up to open search, and swipe left/right/down for seeing the widgets instead.

### LINKED ISSUE #1212: Option to hide search bar completely
Currently you can only make the search bar be:
 - invisible
 - visible with a background
 - visible with transparency

However, if the searchbar is placed at the bottom of the screen while it is invisible, it still adds spacing between the dock/widgets or similar. An option to hide it completely until the drawer is open would solve this.

### LINKED ISSUE #1405: Snap on scroll isn't working for the widgets page.
When you swipe down slightly and release, the homepage slides back up into place. But if you swipe up slightly and release, the screen stops partially between the homepage and the widgets page.

https://github.com/user-attachments/assets/7b6acc51-9e99-4d74-b280-61bb7709cc94

MM2-0:
This PR rewrites the fundamental layout of the launcher.

Goals:
- One scaffold for every configuration; now more duplicate code between pager, pull down and assistant scaffolds
- More flexible configuration, no more limitations,  which component can be placed where 
  - closes #725
  - closes #993
  - closes #1164
- More natural gestures: currently, swipe gestures are handled differently depending on whether they open the search, or widgets, or whether they perform any other action. With this PR, all gestures are handled the same way, resulting in more naturally feeling gestures
  - closes #1003
- Finally hide the search bar completely, when the hidden style is selected
  - closes #57 
  - closes #866 
  - closes #1212
- Add support for predictive back animations

This would also close #1405 and close #1360

TODO / known issues:
- [ ] add a way to access clock widget settings
- [x] implement non-height filling clock widget
- [x] Fix edit widgets padding
- [x] The keyboard sometimes unintentially reopens when overscrolling the search result list


shtrophic:
I've checked it out of curiosity and whether or not the rewrite addresses #1360 as well or not. That doesn't seem to be the case. Might be sensible to address that as well while you are at it?

MM2-0:
> I've checked it out of curiosity and whether or not the rewrite addresses #1360 as well or not. That doesn't seem to be the case. Might be sensible to address that as well while you are at it?

Yes, I'll keep that in mind, hiding system bars isn't implemented yet.

shtrophic:
This closes #381 👍🏼

leekleak:
Fyi there is still an issue where if you expand an item, scroll enough to hide the search bar and close the item, there remains a large gap making the search inaccessible.

https://github.com/user-attachments/assets/6bf5596a-25cd-488d-9a9e-abd1c94885e3

(sorry for the video quality)

leekleak:
Also,

1. I really don't like how when swiping down to search (or to exit search) the app vibrates twice.
2. I really miss the little movement down the search bar currently makes when swiping down to search 🥲

MM2-0:
Thanks for the feedback!

> I really don't like how when swiping down to search (or to exit search) the app vibrates twice.

That's not really new though, is it? It vibrates once when you drag over the activation threshold, and a second time when you release it and activate the search.

> I really miss the little movement down the search bar currently makes when swiping down to search 🥲

I removed it, because the same animation plays when you activate the search when you tap the search bar, and that looked weird. Maybe I'll find a better solution for this problem.

> The dark overlay (underlay?) gives up at the end when scrolling 

I'm aware that, when you overscroll the list, it bounces bit a little which results in this little gap. Have you found a way to permanently have a gap there? Or was this screenshot taken with precise timing to capture the effect?

> Also, maybe it would be better to fade the dark underlay in only once the gesture is finished? The sharp edge looks kinda weird with a bright wallpaper 

I think it looks alright in most cases, but there are a few things I could try: I could try to delay the fade in for a bit (so that it only starts when the drag gesture is over the activation threshold). Or maybe don't make the background part of the slide animation, and just fade it in as the slide progresses. Or maybe add corners, see how that looks.

leekleak:
> That's not really new though, is it? It vibrates once when you drag over the activation threshold, and a second time when you release it and activate the search.

Oh! I hadn't even noticed that! To be honest I think the reason for that is because in the old version the second vibration is much less intense. Also the two vibrations are met with visual feedback - the first one happens when the search bar bottoms out, the second is when it returns to the original position and the app list returns. Without the animation the vibrations are just kinda jarring.

> I'm aware that, when you overscroll the list, it bounces bit a little which results in this little gap. Have you found a way to permanently have a gap there? Or was this screenshot taken with precise timing to capture the effect?

Yeah, that's from the overscroll bounce. Something similar happens when going back to the app list from an app.

> I think it looks alright in most cases, but there are a few things I could try: I could try to delay the fade in for a bit (so that it only starts when the drag gesture is over the activation threshold). Or maybe don't make the background part of the slide animation, and just fade it in as the slide progresses. Or maybe add rounded corners, see how that looks.

My thinking for only fading in the underlay once the widgets have scrolled into place is that it would also help hide the blank space at the top of the widget screen that is reserved for the search bar, which otherwise looks confusing at first glance.

MM2-0:
> @MM2-0 Is a setup like this possible with this pr? I'm really struggling to make the widgets the default screen on nightly 😅

Settings > home screen > widgets on home screen

PR #1430, opened 2025-05-26
Unit conversion "to"

sahilkmishra:
I want to be able to type:

- `14 ft to m`

I find this slightly faster to type than a symbol.

### DIFF
--- a/data/unitconverter/src/main/java/de/mm20/launcher2/unitconverter/UnitConverterRepository.kt
+++ b/data/unitconverter/src/main/java/de/mm20/launcher2/unitconverter/UnitConverterRepository.kt
@@ -80,7 +80,11 @@ internal class UnitConverterRepositoryImpl(
 
 
         // 12cm -> 12 cm so computer sees correct number of tokens
-        val parseableQuery = query.replace(Regex("""([0-9,.:])([a-zA-Z])"""), "$1 $2")
+
+        var parseableQuery = query
+        if (query.matches(Regex("""([0-9,.:])([a-zA-Z])"""))) {
+            parseableQuery = query.replace(Regex("""([0-9,.:])([a-zA-Z])"""), "$1 $2")
+        }
         val valueStr: String
         val unitStr: String
         val targetUnitStr: String?


sahilkmishra:
This contains another change that allows you to set a custom verb in the settings panel of the conversion.
it also adds a change that allows the reader to read the string
12in as 12 in

app/ui/src/main/java/de/mm20/launcher2/ui/settings/unitconverter/UnitConverterSettingsScreenVM.kt
```diff
@@ -42,4 +35,10 @@ class UnitConverterSettingsScreenVM: ViewModel(), KoinComponent {
     val availableUnits = availableConverters.map {
         it.map { converter -> converter.getSupportedUnits() }
     }.shareIn(viewModelScope, SharingStarted.WhileSubscribed(100), 1)
+
+    val customVerb = settings.customVerb
+        .stateIn(viewModelScope, SharingStarted.WhileSubscribed(), "para")
```
MM2-0, line 40: para?

MM2-0:
Can you make the default value translatable? "to" works in English, but other languages might prefer other words

app/ui/src/main/java/de/mm20/launcher2/ui/settings/unitconverter/UnitConverterSettingsScreenVM.kt
sahilkmishra: Spanish/Portuguese, was just testing this out

sahilkmishra:
image
I have added the option to add a 'custom verb' in settings -> Search -> Unit Converter but let me know if this needs to be a custom setting anywhere else

stefnotch:
I'd prefer if there weren't a customisation option for this. The settings are already a hassle to browse through, and it's not obvious what should happen if someone starts configuring options like "mm" (a unit) or "turn into" (multiple words) as the custom verb.

Instead having it be automatically translatable makes a lot of sense to me.

sahilkmishra:
After reading through this again I think I misunderstood what you meant by translatable, I'm away from my computer until the 14th but I'll take a stab at it then. is there a recommended way to get all the translations for this project?

in addition for my use case I'll add a way for the symbol to be ommited aka 17 in cm

PR #1454, opened 2025-06-10
`DevicePoseProvider`: improve `getLocation()`

shtrophic:
The "working variant" of #1344. Turns out that #1344 caused a race-condition for `lastCachedLocation`, since each seperate flow per `LocationItem` would update this global cache, and thus location updates would not propagate to each LocationItem, except the one that "won the race", since `Location.isBetterThan(other)` doesn't return `true` for `this == other`. 

This is why this PR rewrites the logic behind the location caching a bit - I'm submitting this as draft since I'd gladly receive feedback from people here that are able to build a debug APK and test it out in the wild, if location search still works as expected. (#1422 showed that there are some! 😄) Also, I'm unsure if the way I'm using a `ReadWriteLock` here is appropriate, or if there is a more android-y way?


shtrophic:
Did some testing on my phone and seems to work as intended.

PR #1458, opened 2025-06-12
Add background color to blur

leekleak:
Currently the navbar blur has quite a lot of artifacts, especially around the edges of the screen (it also doesn't blur the background). The added background fixes the artifacts, however it also ignores the background, so it's kinda a sidegrade. Still, I think it look a bit less annoying.


MM2-0:
Can you send before and after screenshots? The blur effect looks fine on my device (Pixel 8, Android 16), I haven't noticed any artifacts so far, so it's not clear to me what this does.

leekleak:
Before:


After:


Right now as the text nears the edge of the screen it becomes unblurred

PR #1459, opened 2025-06-12
Reduce calendar refreshes

leekleak:
Currently the calendar widget logic pushes retrieved events as soon as possible instead of aggregating the results and pushing them at once. Since it's impossible for all results to come back at the same time, for the few frames that it takes for all the results to come back the widget bops around.

How that issue looks right now: 
https://github.com/user-attachments/assets/929d0eb4-fab7-4e5a-b095-a0d6070832c7

By waiting for all the results to come in we can avoid that.

Also, since this happens every time you go home/press the home button, I'd like to imagine that this also saves a couple of wasted redraws :Dd

### DIFF
--- a/data/calendar/src/main/java/de/mm20/launcher2/calendar/CalendarRepository.kt
+++ b/data/calendar/src/main/java/de/mm20/launcher2/calendar/CalendarRepository.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.combineTransform
+import kotlinx.coroutines.flow.debounce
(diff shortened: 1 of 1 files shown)


MM2-0:
With this PR, one slow provider is able to slow the entire calendar search down. For example, if you have a calendar plugin provider, that fetches calendar events from a remote server, and that server is slow for some reason, you will have to wait for that server to timeout before you can see your local events.

I believe a better approach would be to use something like [debounce](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/debounce.html)

leekleak:
> I believe a better approach would be to use something like [debounce](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/debounce.html)

Ooooh, that's pretty cool.

Issue #1461, opened 2025-06-14
"Grid & icons” —> "Grid and icons" and "Backup & Restore" —> "Backup and restore"


ghost:
https://m3.material.io/foundations/content-design/style-guide/grammar-and-punctuation#5ed1c04f-6f41-4a5d-8315-a6f786488b4f

PR #1545, opened 2025-07-18
Update strings.xml

jwkwshjsjsj:
Goals:
- Improve consistency
- Remove some "unnecessary" periods
- Modifying some other strings to try to improve them

You can modify the string identifiers if you want to make them be more related to the string values in this pull request.

You can edit this pull request's commits if there is a change you want to make.

Closes https://github.com/MM2-0/Kvaesitso/issues/1461.

### DIFF
--- a/core/i18n/src/main/res/values/strings.xml
+++ b/core/i18n/src/main/res/values/strings.xml
@@ -137,11 +137,11 @@
     <!-- Excel spreadsheet -->
     <string name="file_type_spreadsheet">Spreadsheet</string>
     <!-- Audio file -->
-    <string name="file_type_music">Music file</string>
+    <string name="file_type_music">Audio</string>
     <!-- Video file -->
     <string name="file_type_video">Video</string>
-    <!-- Picture file -->
-    <string name="file_type_image">Picture</string>
+    <!-- Image file -->
+    <string name="file_type_image">Image</string>
     <!-- Generic file without file extension -->
     <string name="file_type_none">File</string>
     <!-- Powerpoint presentation -->
@@ -222,14 +222,14 @@
     <string name="menu_edit_widgets">Edit widgets</string>
     <string name="widget_name_weather">Weather</string>
     <string name="widget_name_calendar">Calendar</string>
-    <string name="widget_name_music">Music</string>
+    <string name="widget_name_music">Media</string>
     <string name="widget_name_favorites">Favorites</string>
     <string name="widget_name_notes">Note</string>
-    <string name="widget_name_unknown">Unknown app widget</string>
+    <string name="widget_name_unknown">Unknown</string>
     <string name="widget_add_widget">Add widget</string>
     <string name="widget_removed">Widget removed</string>
     <!-- Add a third party widget (=a standard Android app widget) -->
-    <string name="notes_widget_placeholder">Write a note…</string>
+    <string name="notes_widget_placeholder">Write a note</string>
     <string name="notes_widget_action_new">New note</string>
     <!-- Save / export the current note content as markdown file -->
     <string name="notes_widget_action_save">Save</string>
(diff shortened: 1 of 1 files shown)


core/i18n/src/main/res/values/strings.xml
```diff
@@ -695,7 +695,7 @@
     <string name="preference_breezyweather_integration_instructions">To use Breezy Weather as weather provider:\n\n1. Go to Breezy Weather &gt; Settings &gt; Widgets &amp; Live wallpaper &gt; Send Gadgetbridge data &gt; Enable %1$s\n\n2. In %1$s, select Breezy Weather as weather provider</string>
     <string name="preference_launch_breezyweather_app">Open Breezy Weather</string>
     <string name="preference_contacts_call_on_tap">Tap to call</string>
-    <string name="preference_contacts_call_on_tap_summary">Immediately start a call when tapping a phone number</string>
+    <string name="preference_contacts_call_on_tap_summary">Start a call when tapping a phone number</string>
```
MM2-0, line 698: I'm not sure if this is clear enough. Tapping a phone number always starts a call, but with this preference disabled, you have to confirm it first, whereas, if you enable the preference, a call is started without confirmation.

core/i18n/src/main/res/values/strings.xml
```diff
@@ -747,24 +747,24 @@
     <string name="create_search_action_type_intent">Custom intent</string>
     <string name="create_search_action_pick_app">Pick an app to search:</string>
     <string name="create_search_action_website_url">Enter the address of the website:</string>
-    <string name="create_search_action_website_invalid_url">The given website cannot automatically be imported as a web search. You can try a different website or enter the required data manually in the next step.</string>
+    <string name="create_search_action_website_invalid_url">The given website cannot automatically be imported as a web search. You can try a different website or enter the required data manually in the next step</string>
```
MM2-0, line 750: I believe a trailing period is necessary when there are multiple sentences?

core/i18n/src/main/res/values/strings.xml
```diff
@@ -403,19 +403,19 @@
...
+    <string name="missing_permission_calendar_widget">Calendar permission is required to display upcoming events here</string>
     <!-- Missing calendar permission, used in calendar search results -->
-    <string name="missing_permission_calendar_search">Grant calendar permission to search your calendar.</string>
+    <string name="missing_permission_calendar_search">Calendar permission is required to search your calendar</string>
     <!-- Missing location permission, used in location search results -->
-    <string name="missing_permission_location_search">Grant location permission to search nearby locations.</string>
+    <string name="missing_permission_location_search">Location permission is required to search nearby locations</string>
     <!-- Missing files permission, used in files search results -->
-    <string name="missing_permission_files_search">Grant storage permission to search photos, media and document on this device.</string>
+    <string name="missing_permission_files_search">Storage permission is required to search audio, documents, images, and videos</string>
```
MM2-0, line 416: All file types are searches, as long as they aren't stored in private directories. Documents, photos, and videos are just the most common types. In file search settings, it says "Search documents, photos, and other files stored on this device", so maybe we shold use that for consistency.

jwkwshjsjsj: > All file types are searches, as long as they aren't stored in private directories. Documents, photos, and videos are just the most common types. In file search settings, it says "Search documents, photos, and other files stored on this device", so maybe we shold use that for consistency.

Isn't the storage permission only for scanning regular media files? Are you sure you're not confusing this translation with the "All files access" translation?

core/i18n/src/main/res/values/strings.xml
MM2-0: It's a bit long, but yeah, that would be a clear summary.

core/i18n/src/main/res/values/strings.xml
MM2-0: Both permissions grant access to the same thing, they just have different names, because they changed the permission model in Android 10. But the actual file search implementation is the same on all Android versions.

ghost:
@MM2-0 Can you make the changes you want to this pull request and merge it?

PR #1560, opened 2025-07-23
Locations: reduce polls

### LINKED ISSUE #1526: High battery usage 1.37.0
Since the new update I have very high battery usage of the launcher. 

There is also a weird behavior where the app language switches between English and Dutch (my phone is set to English (Netherlands)). I tried clearing data and reinstalling the app but didn't seem to have an affect.

Samsung S21 Android 15 (oneui 7.0)

shtrophic:
closes #1526, or shall I say, reduces the amount of time that the device looks for locations when searching, and therefore (possibly) reduces battery usage.


shtrophic:
I think you forgot to invert the early return condition when moving it back to LocationsRepository

MM2-0:
> I think you forgot to invert the early return condition when moving it back to LocationsRepository

Ah, yes 🤦‍♀️

Thanks!

PR #1571, opened 2025-07-28
Rework online results & filters

leekleak:
I feel that the way online results are handled currently create a lot of confusion for new users, so I tried to fix that.

Changes:
- Merged the filter system with the online search system and made it so filters do not filter out existing results but are more like extra buttons to ask for a certain type of search to commence (places, calendar, etc.)
- Reshuffled settings to reflect the changes.

https://github.com/user-attachments/assets/b2f79065-763e-4654-ab20-967289de72f2


leekleak:
P.S. Is it just me or are filters really really useless??

MM2-0:
> P.S. Is it just me or are filters really really useless??

I'm not sure. I don't really use them either, but the intended use is to temporarily enable certain search categories.

The online filter exists for privacy reasons, and to avoid unnecessary server load. There is no point in wasting OSM server resources when you just want to start an app, and you also don't want to find out that everything you've typed into the search bar over the last year has been sent to multiple third party servers.

leekleak:
Yeah that makes sense.

The reason I asked about filters is that I think it they would be much more useful if instead of being filters for all possible results, they'd be buttons to specifically ask for a certain type of result.

That's actually what I thought they were at first. (I thought you had to press the Locations filter to get location results. That wasn't the case.)

If certain searches required a button you could both save on bandwith, declutter the search results and make the filter bar actually useful.

shtrophic:
I think the location results especially are in a weird limbo that consists of the possibility that a hypothetical plugin may return location results without querying the network, which is why both location and online results need to be enabled for OSM to work in particular.

The way you added the online results button to the search bar makes sense to me though - after all, it is not as much a "filter" as the others - also visible currently in the filter bar itself with its separation.

> P.S. Is it just me or are filters really really useless??

Location searches poll GPS which requires additional battery, so that should be user-selectable and not the default behaviour when searching.

leekleak:
> I think the location results especially are in a weird limbo that consists of the possibility that a hypothetical plugin may return location results without querying the network, which is why both location and online results need to be enabled for OSM to work in particular.

Ehh, in my opinion that's not really a problem though? The scenario doesn't just require a theoretical offline map plugin (that doesn't exist), it would also require the user to select **both** the OSM and the offline plugin at the same time. Which wouldn't make sense.

> Location searches poll GPS which requires additional battery, so that should be user-selectable and not the default behaviour when searching.

That's how it works in the branch now :Dd. I've separated the online results so that the app only queries for places when you ask it instead of running all online searches at once.

leekleak:
Also I'd appreciate any and all testing. I don't think the branch is really stable, but the general idea works (I updated the issue text/video to reflect the changes)

leekleak:
ugh this will be impossible to review :(

leekleak:
also, there seems to be a bug (in the launcher, not the pr) where external plugins are not removed from the respective provider lists after they've been deleted.

I haven't looked into this too much, however, I can see that https://github.com/shtrophic/KvaesitsoPublicTransport is still in my location provider list despite having been uninstalled. This probably should be fixed separately though, as the PR is already too big.

PR #1621, opened 2025-09-10
Feature: Allow hiding private space

### LINKED ISSUE #1109: Support fully hiding "Private Space"
As of Android 15, there is a new private space feature included. Whilst this works nicely with Kvaesitso for the most part, there is no way to completely hide it currently. 

Implanting some way of hiding it, similarly to how Google implements hiding in the default Pixel launcher would be an amazing boost to this launcher.

notrudyyy:
Allows hiding the private space via a settings toggle. A temporary solution until https://issuetracker.google.com/issues/352276244 is resolved.

Closes #1109

I have never worked with Kotlin or Android before so it is very likely that this is a messy way to do this. Would appreciate any feedback!


notrudyyy:
@MM2-0 Any feedback?

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/search/SearchVM.kt
```diff
@@ -257,6 +266,13 @@ class SearchVM : ViewModel(), KoinComponent {
                     }
 
             } else {
+                if (hidePrivateProfile.value) {
+                    val score = ResultScore(
+                        query = query,
+                        primaryFields = listOf("Private Space")
```
MM2-0, line 272: This has to be translatable

MM2-0:
Looks good so far. Some things I've noticed:

- If i unlock the private space, it doesn't automatically activate the private space tab, which could be a bit confusing
- The launcher crashes when I lock the private space again, because the selected tab no longer exists
- The "Personal" tab should be hidden if it's the only tab (if the private space is locked and you don't have a work profile).

PR #1649, opened 2025-10-11
Add 8-bit width support to the binary clock

Druthyn:
Adds a toggle for the binary clock to display each value (hours, minutes, seconds) using 8 bits, instead of a mixture of 4, 5, and 6 bits, depending on the value, and whether 24-hour time is used.

Media (27)

- The option is only available in Default mode, as it seemed antithetical to compacts aims.
- Seconds remain optional

As an implementation note, I'm unsure if I needed to make the changes currently commented out in `app/ui/src/main/java/de/mm20/launcher2/ui/launcher/widgets/clock/WatchFaceSelector.kt`. I thought they were, but it seems to function fine without them.


Druthyn:
@MM2-0 Heya, any feedback?

PR #1726, opened 2025-12-30
Fix: Add back pinned tags to apps widget

nemkimo:
In version `1.39.1-fdroid` the pinned tags does not appear on the apps widget anymore when it is used with the favorites and the edit button is disabled. This change is intended to fix this behavior.

While I was fixing the issue above I found that the `Tags` button on the `Favorites` tab in the configuration of the apps widget is basically the same as the `Compact tags` button on the `Tags` tab. Since the text "Tags" is somewhat misleading, I changed it to "Compact tags".


MM2-0:
Just in time, thanks, I almost forgot. I was just about to release a patch update.

Issue #1671, opened 2025-10-31
Blur no longer works when widgets on home screen are enabled


radiohotline:
I am on version 1.38.0-fdroid and Android 15 (Xiaomi). I updated Kvaesitso today and had to fix a few settings as mine broke. One of the things I had to re-enable was widgets on the home screen, but when it's enabled, the wallpaper does not blur when I scroll. When I disable that setting, blurring works. I did try to turn blur on and off and restart my device, but it didn't help.

PR #1791, opened 2026-01-31
Fix status bar and nav bar scrims to show only blurred wallpaper

roccolord:
Restore behavior of "Widgets on home screen" as in pre 1.37.

- Remove colored background overlay from scrims, showing only blur effect
- Remove backgroundColor tint from hazeEffect
- Remove icon color change when scrims are visible (icons now stay based on wallpaper)
- Add wallpaper blur when scrolled away from top on home screen
- Use explicit height for scrim boxes to ensure proper rendering


roccolord:
This PR fixes https://github.com/MM2-0/Kvaesitso/issues/1671

PR #1795, opened 2026-02-03
Add Russian translations

D-ArtGit:
Add Russian translations


D-ArtGit:
Other translations commited

PR #1799, opened 2026-02-07
Added "always show battery" toggle to clock widget

DemonicSavage:
Probably not the best way to do it, but this is good for people like me who like hiding the status bar.

Not my proudest code, I'm not a Kotlin dev at all. If there are better ways of doing what I'm doing please let me know.


DemonicSavage:
What is going on? The original files had no newline at the end, my files also don't. It shouldn't be triggering a diff there

DemonicSavage:
Btw, this should probably be seen as a fairly rough draft. Especially the settings toggle, it's pretty ugly. Also, I'm not so sure about the ideal ranking. 10 seems fine for now though.

MM2-0:
Maybe Battery should be a list preference with 3 options: "Never", "When low or charging" and "always"

DemonicSavage:
Makes sense, done.

DemonicSavage:
As this changes the current schema, I added the migration code.

Unless I'm missing something, this should probably be all.

PR #1805, opened 2026-02-09
Multiple widget panels

### LINKED ISSUE #1575: [Feature Request] Add the ability to have multiple widget panels/pages
Currently there is only one widgets page, it would be nice to have the ability to create at least 2 separate panels, this way I can have a different layout for when I swipe right vs when I swipe left (I have both set to open widgets drawer).

fmontesino:
* Added a new setting under `Home Screen` to allow users to configure the amount of widget screens (1-4). Defaults to 1 to maintain the current production behaviour.
* Added a new `WidgetScreenTarget` to allow widgets to appear independently.
* Changed the `Gestures` setting to display the number of configured Widget screens.
* Changed the `WidgetsComponent` object to a class to be able to have multiple instances.
* Changed `WidgetsVM` to accept a UUID, so we can have multiple instances with different `parents`.

https://github.com/user-attachments/assets/39f67097-f662-4c16-a798-61885b20b93b

Closes #1575

### DIFF
--- a/data/database/src/main/java/de/mm20/launcher2/database/AppDatabase.kt
+++ b/data/database/src/main/java/de/mm20/launcher2/database/AppDatabase.kt
@@ -39,9 +39,11 @@ import de.mm20.launcher2.database.migrations.Migration_22_23
 import de.mm20.launcher2.database.migrations.Migration_23_24
 import de.mm20.launcher2.database.migrations.Migration_24_25
 import de.mm20.launcher2.database.migrations.Migration_25_26
+import de.mm20.launcher2.database.migrations.Migration_26_27
 import de.mm20.launcher2.database.migrations.Migration_27_28
 import de.mm20.launcher2.database.migrations.Migration_28_29
 import de.mm20.launcher2.database.migrations.Migration_29_30
+import de.mm20.launcher2.database.migrations.Migration_30_31
 import de.mm20.launcher2.database.migrations.Migration_6_7
 import de.mm20.launcher2.database.migrations.Migration_7_8
 import de.mm20.launcher2.database.migrations.Migration_8_9
@@ -65,7 +67,7 @@ import java.util.UUID
         ShapesEntity::class,
         TransparenciesEntity::class,
         TypographyEntity::class,
-    ], version = 30, exportSchema = true
+    ], version = 31, exportSchema = true
 )
 @TypeConverters(ComponentNameConverter::class)
 abstract class AppDatabase : RoomDatabase() {
@@ -168,9 +170,11 @@ abstract class AppDatabase : RoomDatabase() {
                         Migration_23_24(),
                         Migration_24_25(),
                         Migration_25_26(),
+                        Migration_26_27(),
                         Migration_27_28(),
                         Migration_28_29(),
                         Migration_29_30(),
+                        Migration_30_31(),
                     ).build()
             if (_instance == null) _instance = instance
             return instance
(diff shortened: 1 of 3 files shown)


app/ui/src/main/java/de/mm20/launcher2/ui/settings/gestures/GestureSettingsScreen.kt
```diff
@@ -58,15 +60,18 @@ fun GestureSettingsScreen() {
...
-                    "${stringResource(R.string.gesture_action_widgets)} ${index + 1}"
+            // Dynamically add widget screen targets based on user configuration
+            WidgetScreenTarget.getAvailableTargets(widgetScreenCount)
+                .forEachIndexed { index, target ->
+
+                    val label =
+                        if (widgetScreenCount == 1) {
+                            stringResource(R.string.gesture_action_widgets)
+                        } else {
+                            "${stringResource(R.string.gesture_action_widgets)} ${index + 1}"
```
MM2-0: This should be a string resource with a placeholder for the index

MelonHell:
I'm really looking forward to trying this. Can I get a build with this feature?

fmontesino:
> I'm really looking forward to trying this. Can I get a build with this feature?

No problem, boss! 🫡 
[Here you go](https://drive.google.com/file/d/1mAjmj_tqYIwlykW0wQ7imdkC6r2mH7jL/view?usp=sharing)
I had to upload it to gdrive as it exceeds the maximum file size on github, apologies!

hunterwilhelm:
Nice work! It does exactly what it’s supposed to. I double-checked to make sure reducing the widget count doesn't wipe out the page settings when I restore it later, and it’s all good. 🎉

fmontesino:
> Nice work! It does exactly what it’s supposed to. I double-checked to make sure reducing the widget count doesn't wipe out the page settings when I restore it later, and it’s all good. 🎉

Thanks for testing! 🫡 
I'm glad you didn't encounter any issues 😄 

@MM2-0 would you mind having a look, see if we could include this in a future release? 🙏

app/ui/src/main/java/de/mm20/launcher2/ui/launcher/scaffold/ClockAndWidgetsHomeComponent.kt
```diff
@@ -124,6 +126,7 @@ internal object ClockAndWidgetsHomeComponent : ScaffoldComponent() {
                     scope.launch { state.lock(hideSearchBar = true) }
                     editMode = it
                 },
+                parentId = Home.scopeId.toString(),
```
MM2-0: This should better use the same ID as the first widgets page, or else you lose all your widgets when you toggle the "Widgets on home" preference

MM2-0 (changes requested):
Good job, but there are some important database migrations missing:

1. When you update from an older version, you lose all your widgets
2. When you clean install the launcher, you start with an empty widgets page (instead of the default page, weather, media and calendar)

fmontesino:
> Good job, but there are some important database migrations missing:
> 
> 1. When you update from an older version, you lose all your widgets
> 2. When you clean install the launcher, you start with an empty widgets page (instead of the default page, weather, media and calendar)

You're absolutely right - good catches! Apologies, it's my first time delving into your -very clean I must say- codebase, so I didn't properly record all database migration paths.

1. I've fixed the correctly identified issue. Furthermore, there was a version for which updates were crashing (unrelated to my changes), since the migration from 26 onwards was missing. This affected version `1.33.1` in particular. I've fixed it by simply adding an empty migration path.

https://github.com/user-attachments/assets/18d34d6e-d4ae-4856-8af7-8f4aa4877f9c


2. Good point, a fresh install will now provide the widgets in the default scope:


https://github.com/user-attachments/assets/4615c6f3-9da5-444d-b570-c96b267b9339

Many thanks for the thorough review! 
I'm loving this contribution experience 😄 

Hopefully we can get this out soon enough 🤞
it was my parter who suggested I should implement the improvement!

Issue #1830, opened 2026-03-04
Option to hide app version/build number in long-press popup


ridd47:
**Feature Request: Option to hide app version/build number in long-press popup**

**Describe the feature**
When long-pressing an app icon on the home screen, the popup displays the app's version/build number. I'd love an option in Settings to toggle this off.

**Why it would be useful**
For users who prefer a cleaner, minimal UI, the version number feels like unnecessary visual clutter in a context menu that's typically used just for quick actions or to open App Info.

**Suggested implementation**
A toggle somewhere in the appearance or home screen settings, such as:
> *Show app version in long-press menu* — On/Off

Thanks for the great launcher!

snowsadh:
Hi, I would like to work on this issue.

PR #1843, opened 2026-03-21
feat : Add setting to toggle app version visibility in grid popup

snowsadh:
## Description
I added a new setting under `Grid and Icons` to show/hide app version in grid's long-press popup.
It is set to visible by default sticking to the old behavior.
Closes issue #1830

## How
- Added gridAppVersion setting
- Implemented UI toggle
- Conditioned rendering based on setting

## Demo
https://github.com/user-attachments/assets/86534e06-7676-4e5f-a753-bbff92027936

This is my first contribution here, happy to make changes based on further feedback or suggestions.

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/base/ProvideSettings.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/base/ProvideSettings.kt
@@ -17,11 +17,13 @@ import de.mm20.launcher2.preferences.MeasurementSystem
 import de.mm20.launcher2.preferences.TimeFormat
 import de.mm20.launcher2.preferences.ui.GridSettings
 import de.mm20.launcher2.preferences.ui.LocaleSettings
+import de.mm20.launcher2.preferences.ui.SearchUiSettings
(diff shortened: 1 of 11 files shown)


MM2-0:
I think the setting would be better placed on settings > search > apps, since it's not really grid related. And also, if you hide the app version, I think you should also hide the package name since it's even less relevant to the average user.

snowsadh:
That makes sense, thanks for the feedback! I’ll move the setting accordingly and update the behavior.
Quick question, would you prefer a single toggle for both version and package name, or separate options?

MM2-0:
One toggle is enough.

snowsadh:
I removed the previous toggle and implemented the new `hideAppDetails` setting at the suggested location.

Here’s a demo:

https://github.com/user-attachments/assets/8ffa9cac-5011-42d4-a199-ea4f3cc7ba91

Does the current padding look fine, or should it be reduced?

Electromagneticradiation:
Hi, just checking in on this PR after the latest updates. Let me know if there’s anything else you’d like me to adjust. @MM2-0 
Thanks!

PR #1855, opened 2026-04-03
Instant search

ImTheLeviDR:
### Motivation
- Remove the artificial staggered delays that caused search results to appear and then update a moment later so drawer search updates appear immediately.

### Description
- Removed `delay(750)` and `delay(250)` that preceded article and location provider searches in `services/search/src/main/java/de/mm20/launcher2/search/SearchService.kt` and removed the now-unused `kotlinx.coroutines.delay` import so those providers start emitting results without waiting.


shtrophic:
Not sure if it is a good idea to call the overpass API on every keypress...

PR #1923, opened 2026-05-29
Feature: Respect hidden private space option

### LINKED ISSUE #1109: Support fully hiding "Private Space"

Boingboingsplat:
Now that 3rd party launchers have access to reading the hidden private space setting in API 36, I wanted to add proper support for it. This is my first PR for an Android project, so any and all feedback is appreciated.

Closes #1109

This PR primarily does three things:
- Filters the private space profile and apps from search while it's locked and hidden.

- Adds a private space quick action that can lock and unlock the private space, even if it's hidden.

    - I added a coroutine in the SearchVM to watch for when the private space action changes the lock to clear the search query, and change to the Private tab when it's been unlocked. It felt a bit hacky to accomplish it this way, so if there's a better way to implement it I'd love to know. I think this is the only search action that does something within the launcher instead of launching another intent?

- Filters private space profile applications from favorites while it's locked and hidden.

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/component/SearchActionIcon.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/component/SearchActionIcon.kt
@@ -127,5 +127,6 @@ fun getSearchActionIconVector(icon: SearchActionIcon): Int {
         SearchActionIcon.Game -> R.drawable.sports_esports_24px
         SearchActionIcon.Note -> R.drawable.sticky_note_2_24px
         SearchActionIcon.Share -> R.drawable.share_24px
+        SearchActionIcon.PrivateSpace -> R.drawable.encrypted_24px
     }
 }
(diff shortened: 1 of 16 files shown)


c89williams:
Thanks for your work on this! I've been enjoying this launcher but encountered the weird issue where private space is always listed as a tab, even when set to be hidden by the OS. Every other launcher I've tested respect this setting, save for Kvaesitso.

Any chance we can we could get this PR merged? Would love to see it!

c89williams:
I tested out a nightly build which includes this feature merge, and while it works as expected now with private space no longer listed as a tab if it is set to hidden, there seems to be a new bug introduced.

Pressing the "lock private space" button (on the private space tab) with the system setting enabled to hide it when locked crashes Kvaesitso. I suspect this is because it is now trying to render a tab that is no longer there.

Issue #1088, opened 2024-09-19
UI improvements for foldable devices


kaibagley:
Hey everyone, 

I made a discussion: https://github.com/MM2-0/Kvaesitso/discussions/1081, but I think this is more suitable as a feature request/issue.

1. Variable apps-per-row in app drawer: When swapping between the thin front screen and wide unfolded screen, it would be great for the number of apps to respond to the width of the screen. At the moment I have 8, which looks fine for the wide screen, but is very tightly packed for the thin screen. Kvaesitso could just double the number of columns in the drawer if the device is open which might be the simplest way to deal with this.
2. Multiple columns for widgets: When scrolling down to the widgets, Kvaesitso fills the width of the screen with the widget, causing them to display as super wide. I think a good solution to this would be to have 2 columns when the device is open.

Thanks for the fantastic launcher, I wish I could contribute some code to this issue, however I know nothing about app development, and less about Kotlin....

StormPooper:
Another potential option - left screen shows the home screen, right shows the widgets view.

Astrrra:
+1 to this, it's so close to being absolutely perfect for me but having a single setting for how many rows of apps I have regardless of whether it's the inner/outer screen drives me crazy...

wipe-tartly-polar:
Can I also second this request?

shakrunk:
I'd like to take a first pass at point 1 (variable apps-per-row in the app drawer).

Proposed approach: rather than adding a separate "columns when unfolded" setting, treat the existing column count setting as the baseline for a compact-width window, and scale it by the current window's Material width size class (compact/medium/expanded, per the [M3 breakpoints](https://m3.material.io/foundations/layout/applying-layout/window-size-classes)) — doubling it at medium/expanded width. That should cover the "double the columns on the wide screen" behavior with no new settings UI. I'm computing the width class from the window size that's already piped through the compose tree, so no new library dependency is needed.

Scope for a first PR: app drawer / search result grid, favorites edit sheet, tag edit sheet, transparency preview — i.e. everywhere the shared grid-columns setting is consumed. The dock/clock-widget inline favorites row reads the column count independently (outside Compose, in a ViewModel) and isn't covered by this pass — can follow up separately.

Point 2 (multi-column widgets) is a bigger change — the widget column is currently a single vertical list with drag-to-reorder, so multi-column needs reorder-across-columns logic. I'll look at that as a follow-up once the grid part lands, to keep the PR reviewable.

[...]

PR #1978, opened 2026-08-09
Scale app drawer grid columns on wider windows (foldables)

shakrunk:
Scales the app drawer / search grid column count on wider windows, so unfolding a foldable device (or otherwise widening the app window) shows more apps per row without a new setting.

## Approach

The existing grid column count setting is treated as the baseline for a compact-width window (< 600dp) and doubled at medium/expanded Material width size classes (≥ 600dp), per the [M3 window size class breakpoints](https://m3.material.io/foundations/layout/applying-layout/window-size-classes). The width class is derived from the window size already piped through the compose tree (`LocalWindowSize`), so no new dependency is needed.

Since `LauncherActivity` has no `android:configChanges` override, a fold/unfold triggers activity recreation, which naturally refreshes the one-shot window size — no extra reactive plumbing required.

Scaling is applied only at grids that actually span the full window width, via a small `scaledGridColumnCount()` helper:

- the app drawer / search results column (`SearchColumn`)
- the favorites grid on the search screen (`SearchFavorites`)

Surfaces whose width is capped regardless of window size — bottom sheets (max 640dp: edit favorites, edit tag, hidden items, icon picker) — deliberately keep the configured base count, since doubling columns in a container that doesn't grow with the window would only shrink the cells.

## Settings: "Density" instead of a column count

Per review feedback, the slider no longer names a column count. It's labelled **Density**, summarised as "How many apps fit in one row", and runs from **Spacious** to **Dense** with no number shown.

Below the slider is a live preview built from the same structure as one row of the real grid (`GridItem`): a weighted cell with 4dp padding, an icon at the actual configured size rather than shrunk to fit. It therefore crowds or spaces out exactly as the real drawer would at that setting, rather than using an invented gap scheme that would look tidier than reality.

The preview is scaled for the window the settings screen is running in *right now* — folded shows the folded count and spacing, unfolded shows the unfolded one — so the effect of this PR is visible directly in the setting. To make that possible, `SettingsActivity` now provides `LocalWindowSize` from the same `Resources.getSystem().displayMetrics` source `SharedLauncherActivity` already uses.

`scaledColumnCount()` / `currentWindowWidthClass()` are the single source of truth for the scaling factor, so the preview and the real grid can't drift apart.

Supporting changes:

- `SliderPreference` gains optional `summary` and `content` slots. Existing call sites are unaffected.
- The density control and its preview hide while "Show apps in a list" is on, matching the sibling toggles — column count has no effect in list mode.
- The icon-pack preview row in the Icons category is scaled the same way, since it also sits at full width; the icon-pack picker sheet stays unscaled, being width-capped.
- The preview row is marked decorative for accessibility (`clearAndSetSemantics`), so TalkBack doesn't walk up to 12 unlabeled placeholders.
- `preference_grid_column_count` is replaced by `preference_grid_density` + `_summary` / `_spacious` / `_dense`. The old string's existing translations are lost by the rename; the four new strings need translating. Happy to reword if you'd prefer keys that reuse more of the existing wording.

## Screenshots

Galaxy Z Fold3 (SM-F926U1), same install and settings (density setting = 6), fold state toggled via `cmd device_state`:

| Folded — cover display (~317dp, compact): 6 columns | Unfolded — inner display (~674dp, medium): 12 columns |
|---|---|
|  |  |

## Behavior changes to be aware of

- **Existing users on wide screens**: tablets and unfolded foldables currently show the configured column count as-is; after this change they show double. Users who tuned their setting for the wide screen (like the reporter of #1088, with 8 columns unfolded) will want to halve it. If an opt-out toggle (or making this opt-in) is preferred, I'm happy to add one — I didn't want to introduce a new setting without maintainer input.
- **Landscape phones**: with fixed rotation disabled, a phone in landscape is ≥ 600dp wide and also gets doubled columns. I've left this in, since it's the same "the window got wider" case and the density preview makes it visible — but it's easy to gate on fold state instead if you'd rather it stay foldable-only.
- The *default* column count is already width-aware at first launch (`values-w400dp`/`values-w480dp` resolve to 5/6), so wide-screen defaults effectively land at 6 × 2 = 12 columns.
- The setting's label changed from "Number of columns" to "Density", so translations for that entry reset.

## Known gaps / limitations

- The dock/clock-widget inline favorites row reads the column count outside the Compose tree (`AppsWidgetVM`, `FavoritesPartProvider`) and is not scaled, so on a wide window it shows fewer items per row than the drawer.
- `LocalWindowSize` is captured from `Resources.getSystem().displayMetrics` at activity creation, i.e. it reflects the display, not the app window — in split-screen/freeform the doubling could trigger in a window that isn't actually wide. (Pre-existing property of that local, not introduced here.)
- Medium windows (600–840dp) get the same 2× as expanded. This matches M3's canonical 4 → 8 column guidance, but makes cells somewhat denser right at 600dp; it could be softened to a smaller factor for medium if preferred.

Multi-column widget layout (the second half of #1088) is a separate, bigger change — the widget column is currently a single vertical list with drag-to-reorder, so multi-column needs reorder-across-columns logic. Planned as a follow-up PR once this lands, to keep this one reviewable.

Ref #1088


MM2-0:
For the settings slider, I would avoid the term "column count" altogether and replace it with something like "density", with a slider from "spacious" to "dense" and a preview how that affects column count.

PR #1983, opened 2026-08-22
Add icon renderer plugin type

ARR8:
Hello, this is a small change to add custom icon renderers via plugins. I wanted to write a custom plugin renderer for my own purposes. I tested it locally and it works fine. I did use AI to write this, because I don't know Kotlin, but I think it is a small change with limited scope for creativity and the diff looks fine to me. Thanks.

### DIFF
--- a/core/base/src/main/java/de/mm20/launcher2/plugin/IconRendererPluginApi.kt
+++ b/core/base/src/main/java/de/mm20/launcher2/plugin/IconRendererPluginApi.kt
@@ -1,0 +1,44 @@
+package de.mm20.launcher2.plugin
+
+import android.content.ComponentName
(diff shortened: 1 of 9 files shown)


MM2-0:
What would be the use case of this that can't already be solved with icon packs?

ARR8:
I don't really know too much about icon packs, so I don't know how dynamic they can be, but I was planning on using this to apply filters to icons. Specifically, I wanted to turn all my icons shapeless by making the background layers transparent.

ARR8:
I found a couple of apps that generate dynamic icon packs, including background removal (Alembicons, Renkin), so I don't need this anymore. Feel free to close.

PR #1999, opened 2026-09-10
don't trigger the long press gesture in the system gesture area

### LINKED ISSUE #1998: Home screen long-press also fires when press-and-hold on the navigation handle opens the Assistant
### Description

On Android with gesture navigation, pressing and holding the **navigation handle** invokes the system Assistant (Gemini / Circle to Search). Kvaesitso's own home screen long-press gesture fires at the same time, so a single press-and-hold triggers both the system action and the launcher action.

The launcher's long-press handler covers the entire home screen surface, including the band at the bottom edge that the system reserves for its own gestures. There is currently no way to avoid this from settings — the long-press gesture is a single global action with no notion of a region, so the only workaround is to disable it entirely (`Settings → Gestures → Long press → No action`), which gives up the gesture across the whole home screen to avoid a conflict confined to a strip at the bottom.

### Steps to reproduce

1. Use gesture navigation, with press-and-hold on the navigation handle bound to the Assistant (the default on recent Android versions).
2. In Kvaesitso, set `Settings → Gestures → Long press` to any action.
3. From the home screen, press and hold the navigation handle.

**Expected:** only the system Assistant opens.
**Actual:** the Assistant opens _and_ the launcher's long-press action fires.

### Why it happens

The home component's gesture handler is attached to the full surface and never consults the system gesture insets:

https://github.com/MM2-0/Kvaesitso/blob/7fd4a6ae74bae3e078e8f3bfd1c8a21ce6c4fe3a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/scaffold/LauncherScaffold.kt#L1296-L1306

```kotlin
config.homeComponent.Component(
    Modifier
        .fillMaxSize()
        .pointerInput(wallpaperManager, config.doubleTap) {
            detectTapGestures(
                onLongPress = config.longPress?.let {
                    { scope.launch { state.onLongPress() } }
                },
                ...
```

Searching the repository, `systemGestures` and `systemGestureExclusion` do not appear anywhere — so the region the platform marks as belonging to system gestures is never taken into account when deciding whether to accept a long press.

### Suggested fix

`detectTapGestures` already passes the press position to `onLongPress`, and it is currently discarded (the lambda ignores its `Offset` parameter). So the position is available without any new plumbing — the guard is a comparison against the bottom gesture inset.

`LauncherScaffold.kt` also already composes window insets a few hundred lines earlier, so the idiom is established in this file:

https://github.com/MM2-0/Kvaesitso/blob/7fd4a6ae74bae3e078e8f3bfd1c8a21ce6c4fe3a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/scaffold/LauncherScaffold.kt#L979-L982

```kotlin
val systemBarInsets = WindowInsets.displayCutout
    .union(WindowInsets.waterfall)
    .let { if (config.showStatusBar) it.union(WindowInsets.statusBars) else it }
    .let { if (config.showNavBar) it.union(WindowInsets.navigationBars) else it }
```

Ignoring a long press whose `Offset` falls inside `WindowInsets.systemGestures`' bottom inset would resolve the conflict while leaving the gesture working across the rest of the home screen. `WindowInsets.navigationBars` would also work if `systemGestures` turns out to be too aggressive on the left and right edges, where it also covers the back-gesture strips.

Happy to open a PR for this if you'd like it — just say which of the two inset types you'd prefer, and whether you'd want it unconditional or behind a setting.

### Possibly related, in the same block

The `pointerInput` key list is `(wallpaperManager, config.doubleTap)` and does not include `config.longPress`. Since `longPress` is nullable and the block is not restarted when it changes, changing the long-press action in settings may not take effect until the launcher restarts. I have not confirmed this on a device — flagging it only because it is cheap to check while the surrounding code is open, and it may be intentional.

### Environment

- Kvaesitso: 1.40.2
- Device: Pixel 10 Pro
- Navigation mode: gesture navigation

kud:
Fixes #1998.

## What this does

The home screen's long-press handler is attached to the full surface, so a press-and-hold on the navigation handle triggers the launcher's long-press action at the same time as the system Assistant. This ignores a long press whose position falls inside the system gesture inset at the bottom of the screen, leaving the gesture working everywhere else.

```kotlin
onLongPress = config.longPress?.let {
    { offset ->
        if (offset.y < (size.height - systemGestureBottomInset).toFloat()) {
            scope.launch { state.onLongPress() }
        }
    }
},
```

`detectTapGestures` was already passing the press position and it was being discarded, so no new plumbing was needed — just the inset value, read the same way the file already reads its other insets a few hundred lines above.

## Choices I made that you may want to override

**Only the bottom inset.** In #1998 I raised that `systemGestures` also covers the left and right back-gesture strips, which could be too aggressive. Using `getBottom()` alone sidesteps that — the sides are untouched. If you would rather this used `navigationBars`, it is a one-word change.

**Unconditional, not a setting.** A press in the system's own gesture area arguably never belonged to the launcher, so this reads as a bug fix rather than something to configure. Easy to put behind a preference if you disagree.

**One extra key on `pointerInput`.** I added `config.longPress` to the key list, which previously held only `wallpaperManager` and `config.doubleTap`. Without it the block is not restarted when the long-press action changes in settings, so changing it appears not to take effect until the launcher restarts. That is the "possibly related" note from #1998 — it is a separate one-line fix riding along, and I am happy to pull it out into its own PR if you would rather keep this focused.

## Why this is a draft — it is not tested

I do not have an Android build environment set up, so **this has not been compiled or run on a device**. I have only checked it by reading: the types line up, `size` resolves from the enclosing `PointerInputScope`, and the import ordering matches the file.

That is not good enough to merge, and I did not want to present it as if it were. What still needs doing:

- [ ] Compiles
- [ ] Long press still works on the home screen proper
- [ ] Long press on the navigation handle no longer fires the launcher action
- [ ] Behaviour is sane with 3-button navigation, where the bottom gesture inset is 0
- [ ] Nothing regresses for double-tap or tap-to-wallpaper-command

I am opening it as a draft in case the approach is useful to you as a starting point, and because the readme suggests small fixes go straight to a PR. To be straight with you about where this leaves things: I am not in a position to set up an Android build environment for this, so I cannot take it further myself and I would not want it merged on my say-so. Treat it as a report with a suggested patch attached rather than a contribution waiting on review — if it is easier to take the idea and implement it your own way, or to close this and do it in your own time, please do. No attachment to the patch.

### DIFF
--- a/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/scaffold/LauncherScaffold.kt
+++ b/app/ui/src/main/java/de/mm20/launcher2/ui/launcher/scaffold/LauncherScaffold.kt
@@ -42,6 +42,7 @@ import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.statusBars
 import androidx.compose.foundation.layout.statusBarsPadding
+import androidx.compose.foundation.layout.systemGestures
 import androidx.compose.foundation.layout.union
 import androidx.compose.foundation.layout.waterfall
 import androidx.compose.foundation.layout.widthIn
@@ -1289,19 +1290,31 @@ internal fun LauncherScaffold(
                 bottom = filterBarHeight
             )
 
+            val systemGestureBottomInset =
+                WindowInsets.systemGestures.getBottom(LocalDensity.current)
+
             CompositionLocalProvider(
                 LocalScaffoldPage provides ScaffoldPage.Home,
             ) {
                 config.homeComponent.Component(
                     Modifier
                         .fillMaxSize()
-                        .pointerInput(wallpaperManager, config.doubleTap) {
+                        .pointerInput(
+                            wallpaperManager,
+                            config.doubleTap,
+                            config.longPress,
+                            systemGestureBottomInset,
+                        ) {
                             detectTapGestures(
                                 onDoubleTap = config.doubleTap?.let {
                                     { scope.launch { state.onDoubleTap() } }
                                 },
                                 onLongPress = config.longPress?.let {
-                                    { scope.launch { state.onLongPress() } }
+                                    { offset ->
+                                        if (offset.y < (size.height - systemGestureBottomInset).toFloat()) {
+                                            scope.launch { state.onLongPress() }
+                                        }
+                                    }
                                 },
                                 onTap = {
                                     wallpaperManager.sendWallpaperCommand(


kud:
@MM2-0 I've just tried tonight and it seems like my patch did not work. Need to have a better look.

